PR #2, opened 2016-02-16
rng-tools: fix build with clang

kergoth:
This fixes 'error: fields must have a constant size' by allocating it with the
given buffer size dynamically.


kergoth:
This has only limited testing, as I don't have a board with a tpm handy, but it does compile at least :)

kergoth:
I'll update it to check for malloc failure and exit.

PR #18, opened 2016-05-10
New nonclangables

dankm:
openjdk supposedly supports clang, but I haven't been able to make it work. I also experienced failures from ncurses and krb5, though I expect those are fixed by my pull request #17.


conf/nonclangable.conf
```diff
@@ -38,6 +38,8 @@ TOOLCHAIN_pn-libtool = "gcc"
 TOOLCHAIN_pn-libvorbis = "gcc"
 TOOLCHAIN_pn-mtools = "gcc"
 TOOLCHAIN_pn-openssl = "gcc"
+TOOLCHAIN_pn-openjdk-8 = "gcc"
+TOOLCHAIN_pn-openjre-8 = "gcc"
```
kraj: would be good to add a short comment on why it fails for future reminder

conf/nonclangable.conf
```diff
@@ -30,6 +30,7 @@ TOOLCHAIN_pn-gnu-efi = "gcc"
 TOOLCHAIN_pn-grub = "gcc"
 TOOLCHAIN_pn-grub-efi = "gcc"
 TOOLCHAIN_pn-gummiboot = "gcc"
+TOOLCHAIN_pn-krb5 = "gcc"
```
kraj: some comment on why this fails

conf/nonclangable.conf
```diff
@@ -38,6 +38,7 @@ TOOLCHAIN_pn-libgfortran = "gcc"
 TOOLCHAIN_pn-libtool = "gcc"
 TOOLCHAIN_pn-libvorbis = "gcc"
 TOOLCHAIN_pn-mtools = "gcc"
+TOOLCHAIN_pn-ncurses = "gcc"
```
kraj: this works ok here. So we need to find why it fails for you

kraj:
Please check if this still is an issue

dankm:
Sure thing. This is really a reabase of commits I made long ago before I backported the -I= stuff. It was failing on armv7.

dankm:
Not being sure if it was still an issue is why this and the backports were separate PRs.

dankm:
I found out why krb5 and ncurses failed. ncurses' configure script strips everything past the first word in $CC, and since my tuning was set to armv7ahf without any of the parameters in $CC clang defaulted to assuming soft-float. Some checks failed, causing the build to fail. I fixed it in this pull request.

Clang erroneously finds an uninitialized variable in krb5. It's fixed upstream in 1.13.4, while meta-openembedded has 1.13.2. I believe the real fix is to update meta-openembedded, but I have a bbappend that fixes it in my master branch if you're curious.

recipes-core/ncurses/ncurses_%.bbappend
```diff
@@ -0,0 +1 @@
+CACHED_CONFIGUREVARS_append_toolchain-clang = " cf_cv_prog_CC_c_o=yes cf_cv_prog_CXX_c_o=yes"
```
kraj: Is ncurses fix still needed ?

dankm: it is on hard-float arm. Anywhere else I've tested it works fine without the ncurses fix.

kraj: > On May 17, 2016, at 9:58 AM, dankm <notifications@github.com> wrote:
> 
> In recipes-core/ncurses/ncurses_%.bbappend <https://github.com/kraj/meta-clang/pull/18#discussion_r63561490>:
> 
> > @@ -0,0 +1 @@
> > +CACHED_CONFIGUREVARS_append_toolchain-clang = " cf_cv_prog_CC_c_o=yes cf_cv_prog_CXX_c_o=yes"
> it is on hard-float arm. Anywhere else I've tested it works fine without the ncurses fix.
> 
> 

I build regularly for rpi3 and haven’t seen this issue. whats best way to reproduce it ?

dankm: I think I narrowed it down further. My host is running CentOS 7 64 bit, I'm testing with MACHINE=wandboard, including oe-core, meta-oe, meta-clang, meta-fsl-arm, and meta-fsl-arm-extra. I *do not* have glibc-devel.i686 installed. With that setup I see in the build log:

configure:2524: checking whether arm-poky-linux-gnueabi-clang understands -c and -o together
configure:2540: arm-poky-linux-gnueabi-clang -c conftest.c -o conftest2.o >&5
In file included from conftest.c:1:
In file included from /usr/include/stdio.h:27:
In file included from /usr/include/features.h:399:
/usr/include/gnu/stubs.h:7:11: fatal error: 'gnu/stubs-32.h' file not found
# include <gnu/stubs-32.h>
          ^
1 error generated.
configure:2543: $? = 1
configure:2562: result: no

As you can see, it's checking the *host* for gnu/stubs-32.h. It's not there on my host because I don't have glibc-devel.i686 installed. If I install glibc-devel.i686 the build succeeds without this change. I'll update the commit message with my new discoveries.

kraj:
Can you update this pull request I think the nonclangable change is still good.

PR #30, opened 2017-03-20
llvm config fixes

martingkelly:
Hi,

This patch series fixes a number of issues I hit when trying to use the
meta-clang llvm-config to cross-compile a project that uses CMake, calling
into llvm-config, to determine compile flags. With the series applied, the
cross compile successfully builds and runs. Please let me know if you have
any concerns or if there is something I missed, and I will fix it up.

### DIFF
--- a/recipes-devtools/clang/clang.inc
+++ b/recipes-devtools/clang/clang.inc
@@ -7,8 +7,10 @@ LLVM_GIT_PROTOCOL ?= "git"
 #
 # Uncomment below to enable master version of clang/llvm
 #
-BASEPV = "4.0.0"
-PV = "${BASEPV}"
+MAJOR_VER="4"
+MINOR_VER="0"
+PATCH_VER="0"
+PV = "${MAJOR_VER}.${MINOR_VER}.${PATCH_VER}"
 BRANCH = "release_40"
 SRCREV_llvm = "4423e351176a92975739dd4ea43c2ff5877236ae"
 SRCREV_clang = "559aa046fe3260d8640791f2249d7b0d458b5700"
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -10,6 +10,11 @@ require clang.inc
 
 PV .= "+git${SRCPV}"
 
+#
+# Default to building only required targets (user overridable).
+#
+LLVM_TARGETS_TO_BUILD ?= "${@get_clang_host_arch(bb, d)};${@get_clang_target_arch(bb, d)}"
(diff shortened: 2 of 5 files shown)


recipes-devtools/clang/clang_git.bb
```diff
@@ -74,10 +81,10 @@ EXTRA_OECMAKE="-DLLVM_ENABLE_RTTI=True \
 "
 
 EXTRA_OECMAKE_append_class-native = "\
-               -DLLVM_TARGETS_TO_BUILD='AArch64;ARM;Mips;PowerPC;X86' \
+               -DLLVM_TARGETS_TO_BUILD="${@get_clang_host_arch(bb, d)};${@get_clang_target_arch(bb, d)}" \
 "
 EXTRA_OECMAKE_append_class-nativesdk = "\
-               -DLLVM_TARGETS_TO_BUILD='AArch64;ARM;Mips;PowerPC;X86' \
+               -DLLVM_TARGETS_TO_BUILD="${@get_clang_host_arch(bb, d)};${@get_clang_target_arch(bb, d)}" \
```
kraj: There is a advantage of building all backends once for all, since I have to test all architectures, it just means that I only build it once ( clang-native ) all other arches then can reuse it, so overall its a big saver for me. I understand that it may be a waste of time for someone who has to deal with limited set of targets.

Could you instead make the LLVM_TARGET_TO_BUILD such that it can be overridden either from config space or via bbapend ?

Then we can have a default value pointing to limited arches as default and I can override it with superset.

recipes-devtools/clang/clang.inc
```diff
@@ -7,7 +7,10 @@ LLVM_GIT_PROTOCOL ?= "git"
 #
 # Uncomment below to enable master version of clang/llvm
 #
-BASEPV = "4.0.0"
+MAJOR_VER="4"
+MINOR_VER="0"
+PATCH_VER="0"
+BASEPV = "${MAJOR_VER}.${MINOR_VER}.${PATCH_VER}"
```
kraj: May be we dont need BASEPV now. Just use the MAJOR MINOR PATCH to achieve the same goal throughout.

recipes-devtools/clang/clang_git.bb
martingkelly: Agreed; I pushed a new version that does this.

recipes-devtools/clang/clang.inc
martingkelly: Agreed; I pushed a new version that removes BASEPV.

PR #65, opened 2018-07-09
libcxx: Add libunwind to PACKAGECONFIG

mdavis777:
Allow libunwind to be turned off and on using PACKAGECONFIG
Default to ON


mdavis777:
This change needs more testing.  It works for what I am using, but I haven't tested it yet with recipes that use clang or with libcxx unwind turned on.

PR #75, opened 2018-12-26
recipes-devtools/clang: Drop virtual/libc-for-gcc

alimon:
No longer required see,

http://git.openembedded.org/openembedded-core/commit/?id=122217b421f749b5fef52bea44ad6e04bc8f8d3a

Signed-off-by: Aníbal Limón <anibal.limon@linaro.org>

### DIFF
--- a/recipes-devtools/clang/clang-cross-canadian_git.bb
+++ b/recipes-devtools/clang/clang-cross-canadian_git.bb
@@ -12,7 +12,7 @@ PN = "clang-cross-canadian-${TRANSLATED_TARGET_ARCH}"
 require clang.inc
 inherit cross-canadian
 
-DEPENDS += "nativesdk-clang binutils-cross-canadian-${TRANSLATED_TARGET_ARCH} virtual/${HOST_PREFIX}binutils-crosssdk virtual/nativesdk-${HOST_PREFIX}libc-for-gcc"
+DEPENDS += "nativesdk-clang binutils-cross-canadian-${TRANSLATED_TARGET_ARCH} virtual/${HOST_PREFIX}binutils-crosssdk virtual/nativesdk-libc"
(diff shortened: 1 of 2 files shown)


recipes-devtools/clang/clang-crosssdk_git.bb
```diff
@@ -11,7 +11,7 @@ PN = "clang-crosssdk-${TARGET_ARCH}"
 
 require clang.inc
 inherit crosssdk
-DEPENDS += "clang-native nativesdk-clang-glue virtual/${TARGET_PREFIX}binutils-crosssdk virtual/nativesdk-${TARGET_PREFIX}libc-for-gcc"
+DEPENDS += "clang-native nativesdk-clang-glue virtual/${TARGET_PREFIX}binutils-crosssdk"
```
kraj: same here replace it with `virtual/nativesdk-libc`

recipes-devtools/clang/clang-cross-canadian_git.bb
```diff
@@ -12,7 +12,7 @@ PN = "clang-cross-canadian-${TRANSLATED_TARGET_ARCH}"
 require clang.inc
 inherit cross-canadian
 
-DEPENDS += "nativesdk-clang binutils-cross-canadian-${TRANSLATED_TARGET_ARCH} virtual/${HOST_PREFIX}binutils-crosssdk virtual/nativesdk-${HOST_PREFIX}libc-for-gcc"
+DEPENDS += "nativesdk-clang binutils-cross-canadian-${TRANSLATED_TARGET_ARCH} virtual/${HOST_PREFIX}binutils-crosssdk"
```
kraj: We can not drop it but replace it with `virtual/nativesdk-libc`

alimon: Updated

PR #83, opened 2019-01-10
clang: Fix binaries built for x86_64 native target (bad interpreter)

romanek-adam:
Assuming the native platform is x86_64 the binaries built by Clang from
clang-native package were properly compiled and linked but could not be
run, e.g.:

> bash: test: No such file or directory

The reason for that was the interpreter set by Clang in the .interp
section of the ELF binary. It was set to /lib/ld-linux-x86-64.so.2
whereas it should be set to /lib64/ld-linux-x86-64.so.2.

This change removes a patch which messes this up.

### DIFF
--- a/recipes-devtools/clang/common.inc
+++ b/recipes-devtools/clang/common.inc
@@ -28,7 +28,6 @@ GCCPIE ??= ""
 
 # Clang patches
 CLANGPATCHES = "\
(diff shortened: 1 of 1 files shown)


kraj:
this is still needed for cross and target clang  just removing it will not help

romanek-adam:
In my current setup I don't need clang-native anymore so I guess this issue can be closed for now.

PR #109, opened 2019-03-14
Fix duplicates symbols when building with lld

MastaG:
Samba: https://svnweb.freebsd.org/ports/branches/2018Q4/net/samba47/files/patch-source3__wscript_build?view=markup&pathrev=481551
Wayland: https://gitlab.freedesktop.org/wayland/wayland/commit/c70fd8a812d74b5ef6fcba1287b409f4f8821e91

### DIFF
--- a/recipes-connectivity/samba/samba_4.8.%.bbappend
+++ b/recipes-connectivity/samba/samba_4.8.%.bbappend
@@ -1,0 +1,3 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI_append += " file://fix_duplicate_pdb_search_init.patch "
--- a/recipes-connectivity/samba/samba_4.8.%.bbappend
+++ b/recipes-connectivity/samba/samba_4.8.%.bbappend
@@ -1,3 +1,3 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
 
-SRC_URI_append += " file://fix_duplicate_pdb_search_init.patch "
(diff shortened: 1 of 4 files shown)


recipes-graphics/wayland/wayland_1.16.%.bbappend
```diff
@@ -0,0 +1,3 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI_append += " file://fix_duplicate_main.patch "
```
kraj: Usually we do not use `+=` and `_append` together, its better to use
`SRC_URI_append = " file://fix_duplicate_main.patch "`

recipes-connectivity/samba/samba_4.8.%.bbappend
```diff
@@ -0,0 +1,3 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI_append += " file://fix_duplicate_pdb_search_init.patch "
```
kraj: we do not use += and _append together, its better to use _append alone

MastaG:
Thank you kraj, I've removed the + signs :)

MastaG:
Hi @kraj 
OE master has switched to wayland 1.17 where the upstream patch has landed.
I'm currently unable to do a PR for a couple of days.
But you can remove the bbappend and patch for wayland 1.16  now :)

kraj:
Yes done with https://github.com/kraj/meta-clang/pull/114

PR #116, opened 2019-04-02
samba: move to dynamic layers

anujm1:
Leads to errors otherwise when networking layer isn't included.

Signed-off-by: Anuj Mittal <anuj.mittal@intel.com>


anujm1:
> we need to add this dynamic layer in layer.conf as well

Sorry I had pushed a wrong version with incomplete fix. Fixed now.

PR #135, opened 2019-08-01
use native-clang compiler when building nativesdk-clang

danieldittmann:
Some cmake based steps when building clang uses BUILD_CC, BUILD_CXX variables,
which points to gcc even if `TOOLCHAIN_class-nativesdk = "clang"`.

This patch sets it to clang if `TOOLCHAIN_class-nativesdk = "clang"` is set.

Signed-off-by: Daniel Dittmann <daniel.dittmann@rohde-schwarz.com>

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -10,6 +10,8 @@ require common-source.inc
 
 INHIBIT_DEFAULT_DEPS = "1"
 
+BUILD_CC_class-nativesdk = "clang"
+BUILD_CXX_class-nativesdk = "clang++"
 BUILD_AR_class-nativesdk = "llvm-ar"
 BUILD_RANLIB_class-nativesdk = "llvm-ranlib"
 BUILD_NM_class-nativesdk = "llvm-nm"


kraj:
One issue I immediately see is that this means we need to have clang installed on the build machine. I am not sure if we want that. We would like to use clang from clang-native

danieldittmann:
> One issue I immediately see is that this means we need to have clang installed on the build machine.
> I am not sure if we want that. We would like to use clang from clang-native
In my tests, no clang was required at the host (where the yocto sdk is built). I run a successful build in a docker container where only the gcc toolchain is available.

In clang_git.bb:
`TOOLCHAIN = "clang"`
`TOOLCHAIN_class-native = "gcc"`
`TOOLCHAIN_class-nativesdk = "clang"`
is set.
AFAIK the gcc is used to build the clang-native, and the clang-native is used to build the nativesdk-clang. That means the meta-clang is building its own clang to compile the clang for the sdk.

danieldittmann:
@kraj Please can you review again? The fix is now limited to the clang recipe only.

kraj:
this looks good

PR #146, opened 2019-08-15
Pass Python3 Exectuable to CMake

### LINKED ISSUE #145: Build Failing with system python executable
My build is failing due to conflicting library and executable versions for Python, for some reason, even though the PATH variable has the path for Python before /usr/bin.

dburnsii:
Even when inheriting Python3native, CMake is picking up the system python executable for some reason, and can cause builds to fail. This fixes that by explicitly providing the path to the executable for CMake to use.

Resolves #145


Lorac:
I was wondering if it was because LLVM_INCLUDE_TESTS was not defined? It defaults to YES and because of that it has to search for a FindPythonInterp

https://github.com/llvm/llvm-project/blob/894abb46f891cba2e0ef581650f27f512a7824b4/clang/CMakeLists.txt#L131

By setting LLVM_INCLUDE_TESTS to off we wouldn't need a dependency on Python.

I'm wondering what's the motivation of activating LLVM_INCLUDE_TESTS?

kraj:
I think it’s acceptable to disable tests for now we are not running build time tests anyway

PR #149, opened 2019-08-16
clang: Pass LLVM_TEMPORARILY_ALLOW_OLD_TOOLCHAIN to stage2 as well

kraj:
Fixes
CMake Error at cmake/modules/CheckCompilerVersion.cmake:105 (message):
  libstdc++ version should be at least 5.1 because LLVM will soon use new C++
  features which your toolchain version doesn't support.  You can temporarily
  opt out using LLVM_TEMPORARILY_ALLOW_OLD_TOOLCHAIN, but very soon your
  toolchain won't be supported.

Signed-off-by: Khem Raj <raj.khem@gmail.com>


kraj:
should fix https://github.com/kraj/meta-clang/issues/147

PR #150, opened 2019-08-17
libcxx: Add libunwind to PROVIDES not only RPROVIDES.

tworaz:
Without this bitbake will still pull standard libunwind package when
building recipes depending on libunwind. If such recipe also happens
to pull libcxx the build will fail when assembling the sysroot because
both libcxx and libunwind provide libunwind.so.

This problem has been observed when building libstd-rs from meta-rust
layer agains musl libc.

Signed-off-by: Piotr Tworek <tworaz@tworaz.net>

### DIFF
--- a/recipes-devtools/clang/libcxx_git.bb
+++ b/recipes-devtools/clang/libcxx_git.bb
@@ -80,7 +80,7 @@ do_install() {
 
 ALLOW_EMPTY_${PN} = "1"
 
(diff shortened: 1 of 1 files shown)


kraj:
thanks @tworaz I had similar patch under test here
https://github.com/YoeDistro/meta-clang/commit/24f35aa94f3a6f465b03d83b693d89f4872192d1

PR #154, opened 2019-08-22
layer: drop sumo support

Lorac:
can't build with sumo because of the dependency on libedit which require native.
libedit doesn't provide BBCLASSEXTEND on sumo.


Lorac:
As we talked face to face, at the conference, I submitted a patch for libedit on thud branch. I did not send it to the sumo branch.

PR #175, opened 2019-09-15
Turn off wchar in editline for native*

jpuhlman:
| In file included from /build/poky/build/tmp/work-shared/llvm-project-source-9.0.0-r0/git/lldb/source/Core/IOHandler.cpp:31:0:
| /build/poky/build/tmp/work-shared/llvm-project-source-9.0.0-r0/git/lldb/include/lldb/Host/Editline.h:36:19:
| fatal error: codecvt: No such file or directory
|  #include <codecvt>
|                    ^

wchar causes codecvt to be included, which does not exist on centos7, which
prevents clang-native from building.

Signed-off-by: Jeremy Puhlman <jpuhlman@mvista.com>


kraj:
hmm seemingly it's due to native gcc being < 5.0 minimum required a version for llvm is 5.1 see

https://llvm.org/docs/GettingStarted.html#software

So I think it would be better that we turn this into a packageconfig options and keep the current default behavior, 

the second option is to use newer GCC on your build host, thats what I do on ubuntu 14.04 I am sure centos7 also provides packages for newer gcc see 

https://ahelpme.com/linux/centos7/how-to-install-gnu-gcc-8-on-centos-7/

jpuhlman:
If meta-clang doesn't support the same hosts that the base yocto release does without system modifications, such as updating the toolchain in the system then, there should probably be a warning and instructions on how to do so.

If you would like this as a packageconfig option, what would you like the option name to be? Honestly, I am not really sure, if clang is written correctly, switching between wchar and otherwise in libedit, should functionally be the same from the users perspective, so doing a packageconfig would seem to me to be overkill, but I could add it if you like.

kraj:
I would like to support as much as possible without losing functionality packageconfig will help in removing it when we do not have distros using gcc < 5 in support anymore

kraj:
applied with pull #176

PR #178, opened 2019-09-18
fixes for multilib and python detection

anujm1:
This makes sure that correct Python (python3native) is detected when compiling target clang and that we're able to compile when multilib is enabled.


anujm1:
I don't see any issue with native stage2 with and without this patch. This should help only the target clang when cross compile is true and llvm tries to build a native version of something for internal use.

PR #203, opened 2019-10-09
clang: make libedit optional

jyong2:
Allows libedit to dependency to be disabled via
PACKAGECONFIG_remove.

Signed-off-by: Yong, Jonathan <jonathan.yong@intel.com>

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -52,7 +52,7 @@ def get_clang_experimental_target_arch(bb, d):
 
 PACKAGECONFIG ??= "compiler-rt libcplusplus shared-libs lldb-wchar \
                    ${@bb.utils.filter('DISTRO_FEATURES', 'thin-lto full-lto', d)} \
-                   rtti eh \
+                   rtti eh libedit \
                    "
 PACKAGECONFIG_class-native = ""
 PACKAGECONFIG_class-nativesdk = "thin-lto"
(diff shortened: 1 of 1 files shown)


jyong2:
Very odd, how is one supposed to make libedit optional if it is still required when disabled? Seems to be passing locally.

kraj:
This is building nativesdk-clang have you tried building that locally

jyong2:
I don't see -DLLVM_ENABLE_LIBEDIT being used anywhere in the logs.
Local nativesdk-clang is passing.

jyong2:
I can't seem to pass the LLVM/LLDB arguments down to the "Configuring NATIVE LLVM" part after doing cleansstate. Any pointers on how to continue?

kraj:
> I can't seem to pass the LLVM/LLDB arguments down to the "Configuring NATIVE LLVM" part after doing cleansstate. Any pointers on how to continue?

lldb has another knob to control it `LLDB_DISABLE_LIBEDIT`. you need to set that along as well

jyong2:
I am currently using:
PACKAGECONFIG[libedit] = "-DLLVM_ENABLE_LIBEDIT=ON -DLLDB_DISABLE_LIBEDIT=0,-DLLVM_ENABLE_LIBEDIT=OFF -DLLDB_DISABLE_LIBEDIT=1,libedit libedit-native"

So LLDB is already set.

It's not passed to the [532/4647] Configuring NATIVE LLVM... stage.

kraj:
Interesting should it be ON/OFF

jyong2:
OK, will try that, I saw 0/1 used when grepping through for LIBEDIT.

jyong2:
Looks like it is still actually broken when reaching "Configuring NATIVE LLVM" if libedit is disabled. Any idea how the cmake command for that stage is generated?

It looks like:
cd ..../work/corei7-64-ese-linux/clang/9.0.0-r0/build/NATIVE &&...../recipe-sysroot-native/usr/bin/cmake

kraj:
merged with changes to fix the build issues via https://github.com/kraj/meta-clang/pull/217

PR #221, opened 2019-11-19
Fix libatomic detection logic

kraj:


anujm1:
This compiles fine for me too. Thanks.

PR #226, opened 2019-11-21
llvm: Check for built-in atomic<double> during configure

mdavis777:
This helps in deciding on linking libatomic, therefore its important to
check for atomic<double> because on x86, clang decides to not use
built-in whereas gcc does, so clangd e.g. links ok when using gcc but
fails when using clang with gcc-runtime on x86

Backport of ba6e545b1d0ee93b4ef698fae2d5a018a062b46a


mdavis777:
Found this issue also exists in warrior  / clang 8.0 so I backported the change

PR #237, opened 2019-12-04
clang: Fix multilib clang compile failing if lib is in the toolchain …

mdavis777:
…path


kraj:
squashed into https://github.com/kraj/meta-clang/pull/239

Issue #271, opened 2020-01-08
[Master] Builds fail with missing __divti3


jpuhlman:
Not sure whats going on here, but I am getting this failure. Any ideas?

| /home/jpuhlman/builds/master/opencgx-x86-generic-64/project/tmp/work/corei7-64-montavista-linux/libcxx/9.0.1-r0/recipe-sysroot-native/usr/bin/x86_64-montavista-linux/x86_64-montavista-linux-ld: projects/libcxx/src/CMakeFiles/cxx_shared.dir/filesystem/operations.cpp.o: in function `std::__1::chrono::__duration_cast<std::__1::chrono::duration<__int128, std::__1::ratio<1l, 1000000000l> >, std::__1::chrono::duration<__int128, std::__1::ratio<1l, 1l> >, std::__1::ratio<1l, 1000000000l>, true, false>::operator()(std::__1::chrono::duration<__int128, std::__1::ratio<1l, 1000000000l> > const&) const':
| /home/jpuhlman/builds/master/opencgx-x86-generic-64/project/tmp/work-shared/llvm-project-source-9.0.1-r0/git/libcxx/include/chrono:876: undefined reference to `__divti3'
| clang-9: error: linker command failed with exit code 1 (use -v to see invocation)
| ninja: build stopped: subcommand failed.

jpuhlman:
Currently building on centos 8.

kraj:
@jpuhlman I have seen this error in past when GCC was used to build libcxx and another case was when using clang with --rtlib=compiler-rt, are there any customizations to such vars done ?

jpuhlman:
Hmmm. I don't see either cases of those in my builds. Ill try backing layers off in my build to see if its something coming from the distro configuration. I don't think I have any configurations that should muck things up.

Here is the line that looks like it fails on.
[...]

PR #273, opened 2020-01-16
libcxx: Correct link failure due to missing __divti3

jpuhlman:
Add libgcc as defined by $CC to link of libc++.

Signed-off-by: Jeremy Puhlman <jpuhlman@mvista.com>


jpuhlman:
This should resolve issue #271

PR #274, opened 2020-01-17
Multilib work

dankm:
Allow multilib configurations to work with clang, libcxx, and compiler-rt. Also use fewer cmake.bbclass overrides at build time.


dankm:
This isn't quite perfect. It works for the main multilib machine, but breaks for the lib32- variants sometimes. Testing a fix now, and new PR will follow.

PR #277, opened 2020-01-23
Use simplier libdir suffix computation

dankm:
An earlier commit added a suffix table to determine the libdir suffix
for multilib targets. This table failed at least in the lib32 case for
x86, where the suffix should have been '', but was instead '32'.

Use the same method as the cmake bbclass instead.

Signed-off-by: Daniel McGregor <daniel.mcgregor@vecima.com>


dankm:
@cosmo0920 Why was the libdir suffix computed the way it was in your commit? Perhaps there's something I'm missing.

cosmo0920:
@dankm Thanks for making simplier implementation.
I've confirmed that your patch works on my multiarch RPi3-64 build environment.
BTW, this changes should be included in warrior branch.

dankm:
@cosmo0920 Thanks for your input. I didn't want to stomp on toes when there could have been a reason I didn't see that it was done that way.

PR #278, opened 2020-01-31
clang,llvm: cmake: Fix configure for packages using find_package() 

ovpanait:
Currently, when a package (i.e. bcc [https://github.com/iovisor/bcc.git])
that depends on LLVM tries to run find_package(LLVM) during do_configure,
it will fail with the following error:

|   The imported target "llvm-tblgen" references the file
|      ".../recipe-sysroot/usr/bin/llvm-tblgen"
|
|   but this file does not exist.  Possible reasons include:
|   * The file was deleted, renamed, or moved to another location.
|   * An install or uninstall procedure did not complete successfully.
|   * The installation package was faulty and contained
|      ".../recipe-sysroot/usr/lib/cmake/LLVMExports.cmake"
|   but not all the files it references.

This is due to the fact that LLVM target binaries will not be present in the
sysroot, only the libraries will. Work around this by not exporting the LLVM
target binaries in LLVMExports cmake files.

Signed-off-by: Ovidiu Panait <ovidiu.panait@windriver.com>

### DIFF
--- a/recipes-devtools/clang/clang/0029-clang-llvm-cmake-Fix-configure-for-packages-using-fi.patch
+++ b/recipes-devtools/clang/clang/0029-clang-llvm-cmake-Fix-configure-for-packages-using-fi.patch
@@ -1,0 +1,138 @@
+From b79837a3b7aa677b44110de4e9e7576fa34a96e3 Mon Sep 17 00:00:00 2001
+From: Ovidiu Panait <ovidiu.panait@windriver.com>
+Date: Fri, 31 Jan 2020 10:56:11 +0200
+Subject: [PATCH] clang,llvm: cmake: Fix configure for packages using
+ find_package()
+
+Currently, when a package (i.e. bcc [https://github.com/iovisor/bcc.git])
+that depends on LLVM/Clang tries to run cmake find_package() during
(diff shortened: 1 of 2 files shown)


ovpanait:
Just realized that clang/cmake/modules/AddClang.cmake has the same issue. Updated the patch.

ovpanait:
So far I have tested it by building bcc [https://github.com/iovisor/bcc] and bpftrace [https://github.com/iovisor/bpftrace.git] which make use of cmake scripts from both clang/cmake/modules and llvm/cmake/modules. 

The results can be replicated by using the following steps:

git clone https://github.com/openembedded/openembedded-core.git
git clone https://github.com/openembedded/bitbake.git
git clone https://github.com/kraj/meta-clang.git
git clone https://github.com/ovpanait/meta-bpf.git

. openembedded-core/oe-init-build-env
bitbake-layers add-layer ../meta-clang/
bitbake-layers add-layer ../meta-bpf/

bitbake bcc
bitbake bpftrace

I will try to find more packages that make use of cmake llvm scripts to further test this.

kraj:
@ovpanait thanks, I will stage it for full world build of Yoe Distro and see if there are any regressions.

PR #284, opened 2020-02-23
Support eBPF tools in meta-clang layer

Leo-Yan:
Since eBPF tool relies on Clang to build eBPF bytecode, this patch set tries to enable the two tools BCC and bpftrace into meta-clang layer.  BCC is dependent on an extra luajit patch for upgrading to 2.1.0-meta3 [1].

This patch set has been tested with qemux86 and qemuarm64.

[1] http://lists.openembedded.org/pipermail/openembedded-devel/2020-February/205070.html

### DIFF
--- a/recipes-devtools/bcc/bcc/0001-Allow-to-build-with-OE-LLVM-cross-compiled-package.patch
+++ b/recipes-devtools/bcc/bcc/0001-Allow-to-build-with-OE-LLVM-cross-compiled-package.patch
@@ -1,0 +1,32 @@
+From 93cf25ba663e68a6a6f4237fbe0ef8349b3f37ef Mon Sep 17 00:00:00 2001
+From: Sumit Garg <sumit.garg@linaro.org>
+Date: Fri, 14 Feb 2020 07:40:11 +0000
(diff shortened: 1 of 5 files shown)


kraj:
@Leo-Yan bcc is failing to build with clang-10 see https://errors.yoctoproject.org/Errors/Details/392565/ would you be able to fix it, otherwise I have to blacklist it.

Leo-Yan:
Hi @kraj,

I checked BCC code, it does support Clang 10 [1]; so the compiling error seems is caused by LLVM_MAJOR_VERSION is not passed properly when use Clang 10.

As I know, I working on meta-clang with version 9.0.1.  Could you let me know how to reproduce the building failure, so I can try to fix it.  Thanks!

[1] https://github.com/iovisor/bcc/blob/master/src/cc/bcc_debug.cc#L204

davidlt:
Clang 10.0 RC2 was merged before BBC. See the history here: https://github.com/kraj/meta-clang/commits/master

Updating your local branch should be enough to update from Clang 9.0.1 to 10.0 RC2.

Leo-Yan:
@davidlt get it, thanks!

PR #286, opened 2020-02-26
BCC: Upgrade to v0.13.0

Leo-Yan:
Since Clang has been upgraded to 10.0.0, it leads to BCC building
failure which is caused by LLVM functions definition mismatching.

This patch upgrades BCC to v0.13.0 so can support Clang 10.0.0.

Signed-off-by: Leo Yan <leo.yan@linaro.org>


kraj:
this still fails btw to build with clang-10
https://errors.yoctoproject.org/Errors/Details/392761/

Leo-Yan:
Reproduced this issue and have updated the patch with an extra line:

COMPATIBLE_HOST = "(x86_64.*|aarch64.*|powerpc64.*)-linux"

Just want to check, so far OE doesn't support s390x building, right?  I searched in meta-openembedded, and don't see any configuration for host s390x.  Otherwise, I should use below line:

COMPATIBLE_HOST = "(x86_64.*|aarch64.*|powerpc64.*|s390.*)-linux"

Very appreciate your time and guidance.

kraj:
we do not support s390 but there is no harm in adding it in COMPATIBLE_HOST, another thing to consider is if it needs clang compiler to build as well or gcc would be fine and just let clang be only a linking item.

PR #289, opened 2020-02-27
[zeus] conf/nonclangable.conf: Always build mesa with gcc

zboszor:
Building mesa (at least version 19.3.4) with clang 9.0.1 leads
to linker errors on x86/x86-64:

"undefined reference to `typeinfo for llvm::raw_pwrite_stream'"
in libvulkan_radeon.so

and

"undefined reference to `typeinfo for llvm::RTDyldMemoryManager'"
in libgallium.a.

It was reported at https://gitlab.freedesktop.org/mesa/mesa/issues/2533

It seems it is related to rtti not being enabled, at least reports
from a few years ago on forums lead to that conclusion.
But enabling rtti for clang in PACKAGECONFIG doesn't help.

Just build mesa with gcc, it fixes the linker errors.

Signed-off-by: Böszörményi Zoltán <zboszor@pr.hu>

### DIFF
--- a/conf/nonclangable.conf
+++ b/conf/nonclangable.conf
@@ -45,10 +45,9 @@ TOOLCHAIN_pn-libssp-nonshared = "gcc"
 TOOLCHAIN_pn-libstd-rs = "gcc"
 TOOLCHAIN_pn-m4_powerpc = "gcc"
 # clang does not have 64bit atomics on mips32
-TOOLCHAIN_pn-mesa_mips = "gcc"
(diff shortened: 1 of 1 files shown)


zboszor:
Overriding mesa with version 19.3.4 in Zeus plus using meta-clang (it's needed by meta-browser) produced the linker errors quoted in the commit message and the gitlab issue also quoted in the commit message.

This is intended only for Zeus, as I haven't tried Yocto master yet.
CLANG/LLVM 10 in master may or may not produce the same linker error.

PR #315, opened 2020-04-26
clang: add lldb, lldb-server, liblldb packages

Lorac:
With lldb split in multiple packages it reduces the size of the `clang` 
package and you can now install lldb, lldb-server and liblldb without h
having the whole clang fat package.

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -215,12 +215,14 @@ do_install_append_class-nativesdk () {
     rm -rf ${D}${datadir}/llvm
 }
 
-PACKAGES =+ "${PN}-libllvm ${PN}-lldb-python libclang"
+PACKAGES =+ "${PN}-libllvm ${PN}-lldb-python libclang lldb lldb-server liblldb"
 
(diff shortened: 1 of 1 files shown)


Lorac:
@kraj Doing a bit more extensive testing, I should've waited.... I got an this issue

```
(lldb) script print "Hello"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
NameError: name 'run_one_line' is not defined
error: python failed attempting to evaluate 'print "Hello"'
```

I added `RDEPENDS_lldb += "readline"` not sure if it's the correct fix.

Lorac:
Actually with 

`RDEPENDS_lldb += "${PN}-lldb-python"`

It seems to work perfectly.

kraj:
> Actually with
> 
> `RDEPENDS_lldb += "${PN}-lldb-python"`
> 
> It seems to work perfectly.

can you use libedit instead you can do so by enabling libedit packageconfig.

davidlt:
This is beneficial as `lldb` is not supported by riscv. I hope to submit my Clang riscv changes after a rebase.

PR #319, opened 2020-04-29
Fix openMP packaging to sit with gnu libgomp

kraj:


davidlt:
Does this mean that `libgomp.so` will not be overwritten by LLVM openmp if both (GNU and LLVM) are installed in rootfs?

kraj:
> Does this mean that `libgomp.so` will not be overwritten by LLVM openmp if both (GNU and LLVM) are installed in rootfs?

that's right. libgomp.so will still come from gcc runtime. However, when something is linked with `clang -fopenmp` it should properly link llvm openmp and use its headers. Currently the problem is that clang can not find omp.h from gcc runtime which is by design I think.

PR #322, opened 2020-05-12
compiler-rt: disable ccache

parr0tr1ver:
It fails to build compiler-rt with ccache. Disable it.

Signed-off-by: Kai Kang <kai.kang@windriver.com>


parr0tr1ver:
@kraj Please ignore this pr for now. I am still working to make ccache enabled. Because every recipe uses native clang may meet this issue. Sorry for inconvenience.

PR #357, opened 2020-08-12
clang: set the default pigz thread

Mingli-Yu:
pigz, which stands for parallel implementation of gzip, is a
fully functional replacement for gzip that exploits multiple
processors and multiple cores to the hilt when compressing data.

do_package uses pigz to compress data to speed up the time
if there is pigz available on the build server.

But for some big package such as clang, there comes below
error during do_package.
gzip: stdout: Cannot allocate memory
tar: TOPDIR/sstate-cache/20/80/sstate:clang:cortexa57-wrs-linux:10.0.1:r0:cortexa57:3:20808f8a746b00ec9470a81057c8595d878cfe8c565cfbcbbaa3ea870528ee34_package.tgz.QF5dss8M: Wrote only 8192 of 10240 bytes
tar: Child returned status 1
tar: Error is not recoverable: exiting now

So set the default pigz thread for clang and the user also can
customize the pigz thread such as PIGZ_pn-clang = "-p 3" for
clang in local.conf to avoid eating so much memory.

Signed-off-by: Mingli Yu <mingli.yu@windriver.com>

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang.inc
+++ b/recipes-devtools/clang/clang.inc
@@ -20,4 +20,7 @@ LLDBMD5SUM = "2e0d44968471fcde980034dbb826bea9"
 
 LLVM_LIBDIR_SUFFIX="${@d.getVar('baselib').replace('lib', '')}"
 
+# set the default pigz thread
+export PIGZ = "-p ${@oe.utils.cpu_count(at_least=2)}"
+
 require common.inc


kraj:
can you instread set `XZ_DEFAULTS` value something like `XZ_DEFAULTS = "--threads=8"`  will that help

Mingli-Yu:
What the XZ_DEFAULTS value do you mean?

Actually I just want to set a default pigz thread which only affect clang build.

kraj:
> What the XZ_DEFAULTS value do you mean?
> 
> Actually I just want to set a default pigz thread which only affect clang build.

pigz says 

-p --processes n
Allow up to n processes (default is the number of online processors)

so what you are setting just for clang could be made global

Mingli-Yu:
Considering it's only set in clang recipe, I think it should only affect clang build. What do you mean "be made global"?

kraj:
Like it is for XZ this should be set for pigz as well and then a custom value can be set for clang if needed because you will have same issue with other packages using llvm eg rust compiler etc

Mingli-Yu:
Understood your concern!
And I even sent the patch as https://patchwork.openembedded.org/patch/175208/, but RP prefer to set for the individual recipe but not want make it global.

PR #391, opened 2020-10-16
Enable building clang for riscv32/64 target

kraj:
---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -35,7 +35,8 @@ def get_clang_arch(bb, d, arch_var):
     elif re.match('aarch64$', a):                      return 'AArch64'
     elif re.match('aarch64_be$', a):                   return 'AArch64'
     elif re.match('mips(isa|)(32|64|)(r6|)(el|)$', a): return 'Mips'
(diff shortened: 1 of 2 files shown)


davidlt:
Looks good, but don't you need to remove LLDB packages from `PACKAGES`? Currently I use:
```
+PACKAGES_remove_riscv32 = "${PN}-lldb-python"
+PACKAGES_remove_riscv32 = "lldb"
+PACKAGES_remove_riscv32 = "lldb-server"
+PACKAGES_remove_riscv32 = "liblldb"
+PACKAGES_remove_riscv64 = "${PN}-lldb-python"
+PACKAGES_remove_riscv64 = "lldb"
+PACKAGES_remove_riscv64 = "lldb-server"
+PACKAGES_remove_riscv64 = "liblldb"
```

You also need to have `fix-clang-driver-gcc-install-path-on-openembedded.patch` to solve Clang driver issues with OpenEmbedded + RISCV.

I have tested that very recently with LLVM/Clang 11 (C/C++, GNU libstdc++v3, libc++, GNU OpenMP and LLVM OpenMP) and all seemed to be working natively. One thing I noticed that if Clang uses GNU OpenMP library in that case it works, but you only get a single thread for some reason (I didn't test that on x86_64).

PR #398, opened 2020-10-21
build shared nativesdk clang

danieldittmann:
The nativesdk gcc is build shared, nativesdk clang should also build shared.
This saves more than 500MB of disk space in sdk.

Signed-off-by: Daniel Dittmann <daniel.dittmann@rohde-schwarz.com>

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -56,7 +56,7 @@ PACKAGECONFIG ??= "compiler-rt libcplusplus shared-libs lldb-wchar \
                    rtti eh libedit \
                    "
 PACKAGECONFIG_class-native = "rtti eh libedit"
-PACKAGECONFIG_class-nativesdk = "rtti eh libedit ${@bb.utils.filter('DISTRO_FEATURES', 'thin-lto full-lto', d)}"
+PACKAGECONFIG_class-nativesdk = "rtti eh libedit shared-libs ${@bb.utils.filter('DISTRO_FEATURES', 'thin-lto full-lto', d)}"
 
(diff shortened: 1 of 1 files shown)


kraj:
how much slow it gets compiling stuff with the SDK compiler after this change ? any data ?

kraj (approved):
lgtm

danieldittmann:
> how much slow it gets compiling stuff with the SDK compiler after this change ? any data ?
I checked up, in my configuration (gcc based SDK with clang as additional compiler included) the SDK takes 2.3GB (with static linked clang) and 1.2GB (with a shared linked clang).
The compile time of the redis-client 1.02 takes 15s (static clang) and 17s (shared clang).
If you think this is significant, I would improve it as separate PACKAGECONFIG option, which the user can activate.

kraj:
That seems like 13% slowdown. How long does this take when compiled with gcc ?

danieldittmann:
> That seems like 13% slowdown. How long does this take when compiled with gcc ?
with gcc it takes 18s

kraj:
@danieldittmann static size of compiler is one time cost but compile time is a time cost you pay every time we compile. I am a bit worries that we are not taking best of the world. I think we need to find middle way. Meaning, we can do shared linking for non compiler tools e.g. lldb clang-scan etc. but see if we can still keep clang and lld like it was.

danieldittmann:
@kraj  Sorry, I hadn't keep in mind that impact. I think I can achieve the behaviour by setting `PACKAGECONFIG_append_pn-nativesdk-clang = " shared-libs"` in local.conf. So we do not need this setting global for all.

kraj:
I am still interested in reducing the size. So your intention is right :) I just want to find a middle path.

danieldittmann:
I try to reduce the size on different ways. I would open an new pr or issue for that.
@kraj Should I revert this change or would you do it?

kraj:
Lets wait how we can improve it.

PR #411, opened 2020-12-14
layer.conf: rename LLVMVERSION to CLANG_LLVMVERSION

sandy-lcq:
In oe-core layer, variable LLVMVERSION is used to set
PREFERRED_VERSION of llvm. Variable LLVMVERSION under
this layer will override value under oe-core layer,
but when llvm verson is different with value set by this
layer. Below warning will appear:

NOTE: preferred version 11.0.1 of llvm not available (for item llvm)
NOTE: versions of llvm available: 10.0.1

Signed-off-by: Changqing Li <changqing.li@windriver.com>

---
### Contributor checklist

### Reviewer Guidelines


kraj:
The idea is exactly to override what’s in oe-core so when meta-clang is added to layers then it should provide all llvm and clang needs in build and override the oe-core version

sandy-lcq:
ok, thanks.  Then I will try to upgrade llvm in oe-core.

PR #429, opened 2021-02-12
libclc: Add recipe

zboszor:
This is a pre-requisite to enable OpenCL in Mesa.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang/0001-Fix-libclc-to-look-for-python3-explicitly.patch
+++ b/recipes-devtools/clang/clang/0001-Fix-libclc-to-look-for-python3-explicitly.patch
@@ -1,0 +1,29 @@
+From 419d8e0602401c813fc5a87b6c593ac0598a445f Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Zolt=C3=A1n=20B=C3=B6sz=C3=B6rm=C3=A9nyi?= <zboszor@pr.hu>
+Date: Fri, 12 Feb 2021 23:56:04 -0500
+Subject: [PATCH] Fix libclc to look for python3 explicitly
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Signed-off-by: Zoltán Böszörményi <zboszor@pr.hu>
+---
+ libclc/CMakeLists.txt | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/libclc/CMakeLists.txt b/libclc/CMakeLists.txt
+index 9472f19..a7908c8 100644
+--- a/libclc/CMakeLists.txt
++++ b/libclc/CMakeLists.txt
+@@ -170,7 +170,7 @@ if( ENABLE_RUNTIME_SUBNORMAL )
+ 		DESTINATION ${CMAKE_INSTALL_DATADIR}/clc )
+ endif()
+ 
+-find_program( PYTHON python )
++find_program( PYTHON python3 )
+ file( TO_CMAKE_PATH ${CMAKE_SOURCE_DIR}/generic/lib/gen_convert.py script_loc )
+ add_custom_command(
+ 	OUTPUT convert.cl
+-- 
+1.8.3.1
+
--- a/recipes-devtools/clang/clang/D95527_against_11.x.patch
+++ b/recipes-devtools/clang/clang/D95527_against_11.x.patch
@@ -1,0 +1,85 @@
+commit b08a140a8fe8d0b0d16a93042b4952d6e34ab913
+Author: Piotr Sobczak <Piotr.Sobczak@amd.com>
+Date:   Wed Jan 27 16:02:49 2021 +0100
+
+    [AMDGPU] Avoid an illegal operand in si-shrink-instructions
+    
+    Before the patch it was possible to trigger a constant bus
+    violation when folding immediates into a shrunk instruction.
+    
+    The patch adds a check to enforce the legality of the new operand.
+    
+    Differential Revision: https://reviews.llvm.org/D95527
+
+diff --git a/llvm/lib/Target/AMDGPU/SIShrinkInstructions.cpp b/llvm/lib/Target/AMDGPU/SIShrinkInstructions.cpp
+index 9c6833a7dab6..6c1b16eddc84 100644
+--- a/llvm/lib/Target/AMDGPU/SIShrinkInstructions.cpp
++++ b/llvm/lib/Target/AMDGPU/SIShrinkInstructions.cpp
+@@ -84,21 +84,23 @@ static bool foldImmediates(MachineInstr &MI, const SIInstrInfo *TII,
+         MachineOperand &MovSrc = Def->getOperand(1);
+         bool ConstantFolded = false;
+ 
+-        if (MovSrc.isImm() && (isInt<32>(MovSrc.getImm()) ||
+-                               isUInt<32>(MovSrc.getImm()))) {
+-          // It's possible to have only one component of a super-reg defined by
+-          // a single mov, so we need to clear any subregister flag.
+-          Src0.setSubReg(0);
+-          Src0.ChangeToImmediate(MovSrc.getImm());
+-          ConstantFolded = true;
+-        } else if (MovSrc.isFI()) {
+-          Src0.setSubReg(0);
+-          Src0.ChangeToFrameIndex(MovSrc.getIndex());
+-          ConstantFolded = true;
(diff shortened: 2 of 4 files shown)


dynamic-layers/meta-python2/recipes-devtools/clang/libclc_git.bb
```diff
@@ -0,0 +1,51 @@
...
+    with open(wrapper_path, 'w') as wrapper_file:
+        wrapper_file.write("""#!/bin/sh
+
+# This file has been generated automatically.
+# It invokes QEMU to run binaries built for the target in the host during the
+# build process.
+
+%s "$@"
+""" % qemu_cmd)
+    os.chmod(wrapper_path, 0o755)
```
kraj, line 44: can we reuse the qemuwrapper from oe-core here ? see `meta/classes/qemu.bbclass` and perhaps see `meta/classes/gtk-doc.bbclass` how it is used

conf/layer.conf
```diff
@@ -17,6 +17,7 @@ BBFILES_DYNAMIC += " \
     openembedded-layer:${LAYERDIR}/dynamic-layers/openembedded-layer/*/*/*.bbappend \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bb \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bbappend \
+    meta-python2:${LAYERDIR}/dynamic-layers/meta-python2/*/*/*.bb \
```
kraj, line 20: can we avoid this please ? py2 is dead and I prefer that we can stay away from it.

kraj (changes requested):

dynamic-layers/meta-python2/recipes-devtools/clang/libclc_git.bb
```diff
@@ -0,0 +1,51 @@
...
+require recipes-devtools/clang/common-source.inc
+
+TOOLCHAIN = "clang"
+
+LIC_FILES_CHKSUM = "file://libclc/LICENSE.TXT;md5=7cc795f6cbb2d801d84336b83c8017db"
+
+inherit cmake pkgconfig pythonnative qemu
+
+DEPENDS_append = " clang-native qemu-native clang"
+
```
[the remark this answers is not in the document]
zboszor: I will remove it, thanks.

dynamic-layers/meta-python2/recipes-devtools/clang/libclc_git.bb
```diff
@@ -0,0 +1,51 @@
...
+OECMAKE_SOURCEPATH = "${S}/libclc"
+
```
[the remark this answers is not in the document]
zboszor: You mean that your own clang, compiler-rt, libcxx and openmp recipe are not good? This line was taken verbatim from those recipes.

conf/layer.conf
zboszor: I will have to add a patch against llvm-project-source so libclc would look for `python3` instead of `python`.
Then it can be moved back to `recipes-devtools/clang`.

dynamic-layers/meta-python2/recipes-devtools/clang/libclc_git.bb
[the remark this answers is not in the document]
kraj: heh good question :) but I think you are right.

conf/layer.conf
kraj: yes such a patch is good and we can upstream it too if not we can maintain it in meta-clang

dynamic-layers/meta-python2/recipes-devtools/clang/libclc_git.bb
zboszor: I will look into this, too.

conf/layer.conf
zboszor: I added the patch, still llvm is compiling on my side but I pushed the changes.
I also added a patch from https://bugs.llvm.org/show_bug.cgi?id=48921 which fixes broken radeonsi shader compilation in some cases, as reported in https://gitlab.freedesktop.org/mesa/mesa/-/issues/4107

dynamic-layers/meta-python2/recipes-devtools/clang/libclc_git.bb
zboszor: The current recipe does use `qemu.bbclass`, it's just adds a python task to create the wrapper as done in `meta-browser/recipes-browser/chromium/chromium-gn.inc`
But I have rewritten it in a shell scriptlet much like in `gtk-doc.bbclass`. The difference is that the wrapper script cannot be in `${B}` because `cmake` pretty much cleans out files it doesn't know about, so it needs to be in `${WORKDIR}`.

kraj (approved):
lgtm

kraj:
btw, I have a clang12 update on a branch and this recipe fails there see https://errors.yoctoproject.org/Errors/Details/570791/
the branch is here
https://github.com/YoeDistro/meta-clang/commits/yoe/mut

Can you see whats going on there ?

zboszor:
> Can you see whats going on there ?

Yes, the same thing that happens with enabling `-Dopencl-spirv` in Mesa.

It needs https://github.com/KhronosGroup/SPIRV-Tools but unlike Mesa where this dependency is optional, it seems to have been made mandatory.

I will create a recipe for spirv-tools. Can this wait until tomorrow? I have this already scheduled at my day job.
Since libclc needs this with LLVM 12, meta-clang can carry it instead of meta-oe or oe-core.

kraj:
http://layers.openembedded.org/layerindex/recipe/152455/ says that spirv-tools is already part of oe-core so all you need is a dependency on it perhaps. I will wait for your inputs.

zboszor:
> http://layers.openembedded.org/layerindex/recipe/152455/ says that spirv-tools is already part of oe-core so all you need is a dependency on it perhaps. I will wait for your inputs.

Nice. Then it's done. Probably spirv-tools-native is what libclc needs because its tools (spirv-as et.al.) will be used during the build.

zboszor:
>     * meta-clang master must break gatesgarth compatibility, or

I meant at the same time llvm 12 is introduced.

zboszor:
There's also another issue I just noticed when trying to build for genericx86: for some reason, `-L/data/OE/sicom-manifest-3.2/tmp-sicom-glibc/work/core2-32-sicom-linux/libclc/11.1.0-r0/recipe-sysroot-native/usr/lib` is passed to the linker command line for the `prepare_builtins` command resulting in a failed build.
The only llvm-config is recipe-sysroot-native/usr/bin/llvm-config and it says:
```
./recipe-sysroot-native/usr/bin/llvm-config --ldflags
-L/data/OE/sicom-manifest-3.2/tmp-sicom-glibc/work/core2-32-sicom-linux/libclc/11.1.0-r0/recipe-sysroot-native/usr/lib   -isystem/data/OE/sicom-manifest-3.2/tmp-sicom-glibc/work/x86_64-linux/clang-native/11.1.0-r0/recipe-sysroot-native/usr/include -O2 -pipe   -L/data/OE/sicom-manifest-3.2/tmp-sicom-glibc/work/x86_64-linux/clang-native/11.1.0-r0/recipe-sysroot-native/usr/lib                         -L/data/OE/sicom-manifest-3.2/tmp-sicom-glibc/work/x86_64-linux/clang-native/11.1.0-r0/recipe-sysroot-native/lib                         -Wl,--enable-new-dtags                         -Wl,-rpath-link,/data/OE/sicom-manifest-3.2/tmp-sicom-glibc/work/x86_64-linux/clang-native/11.1.0-r0/recipe-sysroot-native/usr/lib                         -Wl,-rpath-link,/data/OE/sicom-manifest-3.2/tmp-sicom-glibc/work/x86_64-linux/clang-native/11.1.0-r0/recipe-sysroot-native/lib                         -Wl,-rpath,/data/OE/sicom-manifest-3.2/tmp-sicom-glibc/work/x86_64-linux/clang-native/11.1.0-r0/recipe-sysroot-native/usr/lib                         -Wl,-rpath,/data/OE/sicom-manifest-3.2/tmp-sicom-glibc/work/x86_64-linux/clang-native/11.1.0-r0/recipe-sysroot-native/lib                         -Wl,-O1
```
How can I convince clang to build an llvm-config that points to the `recipe-sysroot` of the actual recipe instead of its own build's original `recipe-sysroot-native`?

zboszor:
@kraj It seems `llvm-common` ships `crossscripts/llvm-config` that does what's expected for target builds but no other recipe adds it via `DEPENDS`. How comes?

PR #432, opened 2021-02-18
libclc: Fix linker issue if build host is the same arch as target

zboszor:
llvm-config from clang-native leaks -L linker options to target builds.
Add llvm-common as build dependency for target builds to fix it.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/libclc_git.bb
+++ b/recipes-devtools/clang/libclc_git.bb
@@ -14,6 +14,7 @@ inherit cmake pkgconfig python3native qemu
 SPIRV_DEP = "${@'spirv-tools spirv-tools-native' if int(d.getVar('MAJOR_VER')) >= 12 else ''}"
 
 DEPENDS_append = " qemu-native clang ${SPIRV_DEP}"
+DEPENDS_append_class-target = " llvm-common"
 
 OECMAKE_SOURCEPATH = "${S}/libclc"
 


recipes-devtools/clang/libclc_git.bb
```diff
@@ -14,6 +14,7 @@ inherit cmake pkgconfig python3native qemu
 SPIRV_DEP = "${@'spirv-tools spirv-tools-native' if int(d.getVar('MAJOR_VER')) >= 12 else ''}"
 
 DEPENDS_append = " qemu-native clang ${SPIRV_DEP}"
+DEPENDS_append_class-target = " llvm-common"
```
kraj, line 17: i wonder if it should br in DEPENDS of clang recipe

zboszor: See my question from the first libclc PR: https://github.com/kraj/meta-clang/pull/429#issuecomment-778755847
No other llvm subproject recipe has it.

zboszor: But sure, the clang recipe should likely merge llvm-common and install llvm-common (and the versioned variant) in the crossscripts directory.

zboszor: Alternative: https://github.com/kraj/meta-clang/pull/433

PR #438, opened 2021-03-04
spirv-llvm-translator: disable ccache

saininav:
Native build is failing, error log:
~/build/tmp/work/x86_64-linux/spirv-llvm-translator-native/12.0.0-r0/git/lib/SPIRV/SPIRVLowerSPIRBlocks.cpp
| /bin/sh: 1: ccache: not found

Signed-off-by: Naveen Saini <naveen.kumar.saini@intel.com>

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/spirv-llvm-translator/spirv-llvm-translator/0001-cmake-allow-to-enable-disable-ccache.patch
+++ b/recipes-devtools/spirv-llvm-translator/spirv-llvm-translator/0001-cmake-allow-to-enable-disable-ccache.patch
@@ -1,0 +1,32 @@
+From 13b679eb76e6877b40fd41b91e71cb172bcf7b26 Mon Sep 17 00:00:00 2001
+From: Naveen Saini <naveen.kumar.saini@intel.com>
+Date: Thu, 4 Mar 2021 11:23:20 +0800
+Subject: [PATCH] cmake: allow to enable/disable ccache
+
+By default ccache is enabled.
+
+Upstream-Status: Submitted [https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/930]
+
+Signed-off-by: Naveen Saini <naveen.kumar.saini@intel.com>
+---
+ CMakeLists.txt | 3 ++-
+ 1 file changed, 2 insertions(+), 1 deletion(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 0ddb55c..c7d0498 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -50,8 +50,9 @@ if(LLVM_SPIRV_BUILD_EXTERNAL)
+ 
+   message(STATUS "Found LLVM: ${LLVM_VERSION}")
+ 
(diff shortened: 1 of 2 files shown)


kraj (changes requested):
it seems you are enabling ccache globally. Can you add

```
CCACHE_DISABLE = "1"
```
to this recipe and see if it helps? if it does then we can avoid maintaining one extra patch

saininav:
It does not help. Native build still fails.

saininav:
If I need to disable ccache only for native, than I should inherit ccache for target build and carry a patch to disable for native build.
By default ccache is not detected for target build.

kraj:
> If I need to disable ccache only for native, than I should inherit ccache for target build and carry a patch to disable for native build.
> By default ccache is not detected for target build.

I see 
is ccache exposed via HOSTTOOLS ?

`bitbake -e <recipe> | grep ^HOSTTOOLS=`

saininav:
No.

HOSTTOOLS="      [ ar as awk basename bash bzip2 cat chgrp chmod chown chrpath cmp comm cp cpio     cpp cut date dd diff diffstat dirname du echo egrep env expand expr false     fgrep file find flock g++ gawk gcc getconf getopt git grep gunzip gzip     head hostname iconv id install ld ldd ln ls make md5sum mkdir mkfifo mknod     mktemp mv nm objcopy objdump od patch perl pr printf pwd     python3 ranlib readelf readlink realpath rm rmdir rpcgen sed seq sh     sha1sum sha224sum sha256sum sha384sum sha512sum     sleep sort split stat strings strip tail tar tee test touch tr true uname     uniq wc wget which xargs  "

saininav:
From native log.do_configure :

-- Found ccache: /usr/bin/ccache

recipes-devtools/spirv-llvm-translator/spirv-llvm-translator/0001-cmake-allow-to-enable-disable-ccache.patch
```diff
@@ -0,0 +1,32 @@
...
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -50,8 +50,9 @@ if(LLVM_SPIRV_BUILD_EXTERNAL)
+ 
+   message(STATUS "Found LLVM: ${LLVM_VERSION}")
+ 
++  option(CCACHE_ALLOWED "allow use of ccache" TRUE)
+   find_program(CCACHE_EXE_FOUND ccache)
+-  if(CCACHE_EXE_FOUND)
++  if(CCACHE_EXE_FOUND AND CCACHE_ALLOWED)
```
kraj, line 26: lets rename this knob to `ENABLE_CCACHE`

saininav: PR already approved ! 
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/930

kraj: nice

Issue #439, opened 2021-03-08
ROCm-CompilerSupport do_configure fails with meta-clang


zboszor:
**Describe the bug**

https://github.com/RadeonOpenCompute/ROCm-CompilerSupport/blob/amd-stg-open/lib/comgr/CMakeLists.txt#L66 uses
```
  find_package(LLD REQUIRED CONFIG)
```
which for target builds fail with
```
| ERROR: Execution of '/home/zozo/sicom-manifest-3.3/tmp-sicom-glibc/work/corei7-64-sicom-linux/amd-comgr/4.0.0-r0/temp/run.do_configure.1199779' failed with exit code 1:
| -- The C compiler identification is Clang 12.0.0
| -- The CXX compiler identification is Clang 12.0.0
| -- Detecting C compiler ABI info
| -- Detecting C compiler ABI info - done
| -- Check for working C compiler: /home/zozo/sicom-manifest-3.3/tmp-sicom-glibc/work/corei7-64-sicom-linux/amd-comgr/4.0.0-r0/recipe-sysroot-native/usr/bin/x86_64-sicom-linux/x86_64-sicom-linux-clang - skipped
| -- Detecting C compile features
| -- Detecting C compile features - done
| -- Detecting CXX compiler ABI info
| -- Detecting CXX compiler ABI info - done
| -- Check for working CXX compiler: /home/zozo/sicom-manifest-3.3/tmp-sicom-glibc/work/corei7-64-sicom-linux/amd-comgr/4.0.0-r0/recipe-sysroot-native/usr/bin/x86_64-sicom-linux/x86_64-sicom-linux-clang++ - skipped
| -- Detecting CXX compile features
| -- Detecting CXX compile features - done
| -- Found ZLIB: /home/zozo/sicom-manifest-3.3/tmp-sicom-glibc/work/corei7-64-sicom-linux/amd-comgr/4.0.0-r0/recipe-sysroot/usr/lib/libz.so (found version "1.2.11")
| -- Found LibXml2: /home/zozo/sicom-manifest-3.3/tmp-sicom-glibc/work/corei7-64-sicom-linux/amd-comgr/4.0.0-r0/recipe-sysroot/usr/lib/libxml2.so (found version "2.9.10")
[...]

PR #440, opened 2021-03-10
clang: Install lld for target builds

zboszor:
LLDTargets.cmake (generated and installed by the lld build)
references lld this way:

add_executable(lld IMPORTED)
set_property(TARGET lld PROPERTY ENABLE_EXPORTS 1)

This way lld can be used by plugins to get their symbols from,
pretty much like PHP modules that resolve their symbols from the
php executable during runtime linking.

At least https://github.com/RadeonOpenCompute/ROCm-CompilerSupport
uses lld's cmake files as opposed to using only llvm and previously
failed with:

|   The imported target "lld" references the file
|
|      ".../tmp-sicom-glibc/work/corei7-64-sicom-linux/amd-comgr/4.0.0-r0/recipe-sysroot/usr/bin/lld"
|
|   but this file does not exist.  Possible reasons include:
|
|   * The file was deleted, renamed, or moved to another location.
|
|   * An install or uninstall procedure did not complete successfully.
|
|   * The installation package was faulty and contained

Extending SYSROOT_DIRS with ${bindir} would break crosscompiling
since CMake found clang/clang++ from recipe-sysroot in this case.
Install lld into sysroot-destdir but nothing else.

Fixes: #439 

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -303,4 +303,7 @@ clang_sysroot_preprocess() {
 	install -d ${SYSROOT_DESTDIR}${bindir_crossscripts}/
 	install -m 0755 ${S}/../llvm-config ${SYSROOT_DESTDIR}${bindir_crossscripts}/
 	ln -sf llvm-config ${SYSROOT_DESTDIR}${bindir_crossscripts}/llvm-config${PV}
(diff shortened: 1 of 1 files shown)


kraj (approved):
lgtm

PR #441, opened 2021-03-10
bcc: link with dynamic libbpf

teknoraver:
Link bcc dynamically against libbpf. This reduces the size of libbcc.so
from 68 MB to 3 MB.
Backport some patches from upstream which are needed for the actual
dynamic linking.

Depends on openembedded/meta-openembedded#320

Signed-off-by: Matteo Croce <mcroce@microsoft.com>

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):

kraj:
I am seeing build failures on aarch64/musl see http://sprunge.us/nAGWNa

teknoraver:
Yes, this is expected as the PR depends on openembedded/meta-openembedded#320

Can you try both?

kraj:
it is using both

teknoraver:
OK, so it should be specific to musl. I tried only x86_64/libc and aarch64/libc6.

If all the other configurations work, we can exclude musl or stick with static linking

kraj:
I think the issue I was facing was different, where a dependency was missing I have sent a patch for that.

Issue #466, opened 2021-05-23
bitbake -c scanview error on debian 10


dl9pf:
**Describe the bug**
When running bitbake -c scanview on a debian 10 host using 
Build Configuration:
BB_VERSION           = "1.50.0"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "universal"
TARGET_SYS           = "x86_64-poky-linux"
MACHINE              = "qemux86-64"
DISTRO               = "poky"
DISTRO_VERSION       = "3.3.1"
TUNE_FEATURES        = "m64 core2"
TARGET_FPU           = ""
meta                 
meta-poky            
meta-yocto-bsp       = "hardknott:05a8aad57ce250b124db16705acec557819905ae"
meta-clang           = "master:d797409435d3b0e9f2859992439989ff1d81e66d"

scanview does not start and errors out with:

Log data follows:
| DEBUG: Executing shell function do_scanview
| ================================================================
| Stating scan-view server at: http://192.168.2.6:8181
| Use Ctrl-C to exit
| ================================================================
| /home/dl9pf/YPDD_STATIC/build/tmp/work/core2-64-poky-linux/busybox/1.33.0-r0/recipe-sysroot-native/usr/bin/scan-view:9: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
|   import imp
| Traceback (most recent call last):
|   File "/home/dl9pf/YPDD_STATIC/build/tmp/work/core2-64-poky-linux/busybox/1.33.0-r0/recipe-sysroot-native/usr/bin/scan-view", line 150, in <module>
|     main()
|   File "/home/dl9pf/YPDD_STATIC/build/tmp/work/core2-64-poky-linux/busybox/1.33.0-r0/recipe-sysroot-native/usr/bin/scan-view", line 147, in main
|     run(port, args, args.root)
|   File "/home/dl9pf/YPDD_STATIC/build/tmp/work/core2-64-poky-linux/busybox/1.33.0-r0/recipe-sysroot-native/usr/bin/scan-view", line 74, in run
|     import ScanView
|   File "/home/dl9pf/YPDD_STATIC/build/tmp/work/core2-64-poky-linux/busybox/1.33.0-r0/recipe-sysroot-native/usr/bin/../share/scan-view/ScanView.py", line 29, in <module>
|     import Reporter
| ModuleNotFoundError: No module named 'Reporter'
| WARNING: Unable to detect that server started.
[...]

PR #467, opened 2021-05-24
Fix for Reporter.py

dl9pf:
See: https://github.com/kraj/meta-clang/issues/466


dl9pf:
Yeez, I need coffee and sleep. Will fix in v2. Copy'pasta bug.

dl9pf:
Here is v2. Build passed.

kraj (approved):

PR #490, opened 2021-08-04
clang: support android runtime

hizukiayaka:
Or I should set up a new RUNTIME condition? Because in Android, only the C++ environment comes from LLVM but the libc is not .
Signed-off-by: Hsia-Jun(Randy) Li <randy.li@synaptics.com>

---

### DIFF
--- a/classes/clang.bbclass
+++ b/classes/clang.bbclass
@@ -91,15 +91,16 @@ def clang_base_deps(d):
     if not d.getVar('INHIBIT_DEFAULT_DEPS', False):
         if not oe.utils.inherits(d, 'allarch') :
             ret = " clang-cross-${TARGET_ARCH} virtual/libc "
+            if (d.getVar('RUNTIME').find('android') != -1):
+                ret += " libcxx"
+                return ret
             if (d.getVar('RUNTIME').find('llvm') != -1):
                 ret += " compiler-rt libcxx"
             elif (d.getVar('COMPILER_RT').find('-rtlib=compiler-rt') != -1):
                 ret += " compiler-rt "
             else:
                 ret += " libgcc "
-            if (d.getVar('RUNTIME').find('llvm') != -1):
-                ret += " libcxx"
-            elif (d.getVar('COMPILER_RT').find('--unwindlib=libunwind') != -1):
+            if (d.getVar('COMPILER_RT').find('--unwindlib=libunwind') != -1):
                 ret += " libcxx "
             elif (d.getVar('LIBCPLUSPLUS').find('-stdlib=libc++') != -1):
                 ret += " libcxx "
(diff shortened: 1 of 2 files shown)


kraj:
> Or I should set up a new RUNTIME condition? Because in Android, only the C++ environment comes from LLVM but the libc is not .
> Signed-off-by: Hsia-Jun(Randy) Li [randy.li@synaptics.com](mailto:randy.li@synaptics.com)

thanks for your contribution. android comes with bionic libc so I would assume that you are setting TCLIBC to bionic and that variable could be used. But perhaps using a third value for RUNTIME is also a good solution.

hizukiayaka:
> > Or I should set up a new RUNTIME condition? Because in Android, only the C++ environment comes from LLVM but the libc is not .
> > Signed-off-by: Hsia-Jun(Randy) Li [randy.li@synaptics.com](mailto:randy.li@synaptics.com)
> 
> thanks for your contribution. android comes with bionic libc so I would assume that you are setting TCLIBC to bionic and that variable could be used. But perhaps using a third value for RUNTIME is also a good solution.

Yes, I did
https://github.com/hizukiayaka/meta-android-ndk/blob/master/conf/distro/include/tcmode-external-android.inc#L4

I was wondering whether I should keep the name `libcxx-external`, although I don't think people would like to use the clang shipped in Android NDK to build a new c++ std library but c++ library is not a part of Android system, maybe your application could use your version.

kraj:
> > > Or I should set up a new RUNTIME condition? Because in Android, only the C++ environment comes from LLVM but the libc is not .
> > > Signed-off-by: Hsia-Jun(Randy) Li [randy.li@synaptics.com](mailto:randy.li@synaptics.com)
> > 
> > 
> > thanks for your contribution. android comes with bionic libc so I would assume that you are setting TCLIBC to bionic and that variable could be used. But perhaps using a third value for RUNTIME is also a good solution.
> 
> Yes, I did
> https://github.com/hizukiayaka/meta-android-ndk/blob/master/conf/distro/include/tcmode-external-android.inc#L4
> 
> I was wondering whether I should keep the name `libcxx-external`, although I don't think people would like to use the clang shipped in Android NDK to build a new c++ std library but c++ library is not a part of Android system, maybe your application could use your version.

I am willing to accept patches to fix existing libcxx to be able to be used for android as well.

hizukiayaka:
I could make libcxx-external provides libcxx and virtual/${TARGET_PREFIX}compilerlibs.
But libcxx is not a virtual package, then libcxx would not be built, what I did here would make dependency information from libcxx be used while we are using libcxx-external.
libcxx-external doesn't have so many dependency.
I am thinking a dynamic layer to solve this problem.

kraj:
> I could make libcxx-external provides libcxx and virtual/${TARGET_PREFIX}compilerlibs. But libcxx is not a virtual package, then libcxx would not be built, what I did here would make dependency information from libcxx be used while we are using libcxx-external. libcxx-external doesn't have so many dependency. I am thinking a dynamic layer to solve this problem.

`libcxx-external` recipe could add `PROVIDES += "libcxx"` and then in local.conf one can set `PREFERRED_PROVIDER_libcxx = "libcxx-external"` logic in `clang.bbclass` adds it to DEPENDS and does not care who provides `libcxx` during build.

kraj (approved):
looks ok to me as such.

PR #506, opened 2021-09-10
Allow building meta-browser/master with meta-clang/gatesgarth

shr-project:
Fixes https://github.com/OSSystems/meta-browser/issues/560


kraj (approved):
gatesgarth is EOL but lets still install it.

PR #513, opened 2021-09-22
clang: use global multilib variants

dankm:
Using regular multilib variants prevents sharing state between machines
with dissimilar multilibs, or non-multilib machines. Since there's no
harm in having extra multilib triples, use the full set for all builds.

Signed-off-by: Dan McGregor <dan.mcgregor@usask.ca>

---
### Contributor checklist

### Reviewer Guidelines


kraj (commented):
this is failing to build some packages. Can you take a look and see why ?

dankm:
Absolutely. The first build failure didn't look like it was related to my change, but the new one in do_testimage looks like something I can try to reproduce.

PR #518, opened 2021-10-05
DEPENDS on virtual/${TARGET_PREFIX}binutils

guillaume-pais-siemens:
```virtual/${TARGET_PREFIX}binutils``` shall be used and not ```binutils-cross-${TARGET_ARCH}```
In the case of an external arm toolchain, ```binutils-cross-aarch64``` does not exist and creates following error:
```
Missing or unbuildable dependency chain was: ['tisdk-base-image', 'packagegroup-core-standalone-sdk-target', 'libcxx-dev', 'compiler-rt', 'clang-cross-aarch64', 'binutils-cross-aarch64']
```
Signed-off-by: Guillaume Pais <guillaume.pais@siemens.com>

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang-cross_git.bb
+++ b/recipes-devtools/clang/clang-cross_git.bb
@@ -11,7 +11,7 @@ PN = "clang-cross-${TARGET_ARCH}"
 require clang.inc
 require common-source.inc
 inherit cross
-DEPENDS += "clang-native binutils-cross-${TARGET_ARCH}"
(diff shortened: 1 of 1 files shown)


kraj:
I think this should be fixed in external toolchain recipe since the naming conventions we use are inline with latest openembedded cross toolchain components

kraj (approved):
LGTM

PR #524, opened 2021-11-01
clang: Backport missing include fix from upstream.

rakuco:
This affects LLVM < 12, so backport directly to dunfell. This adds a missing
include that was breaking the build with recent libstdc++ releases on the
host system (versions 11 and above).


kraj (approved):
lgtm

PR #531, opened 2021-11-22
recipes: Specify branch and protocol in SRC_URI

tewarid:
github has stopped supporting anonymous git protocol and bitbake now
warns about branch not being explicit, master is no more a viable
fallback


kraj (approved):
LGTM

PR #540, opened 2021-12-08
bcc: Add explicit dependency on flex target recipe

b49020:
While building bcc for the target it depends on flex target recipe. So
add that dependency explicity.

Signed-off-by: Sumit Garg <sumit.garg@linaro.org>


kraj (approved):

b49020:
@kraj It looks like the build failure reported isn't related to this PR.

kraj:
> @kraj It looks like the build failure reported isn't related to this PR.

yes

PR #547, opened 2021-12-24
clang: support to compile with out of system-wide Yocto GCC

hongxu-jia:
---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang/0039-support-to-compile-with-out-of-system-wide-GCC.patch
+++ b/recipes-devtools/clang/clang/0039-support-to-compile-with-out-of-system-wide-GCC.patch
@@ -1,0 +1,91 @@
+From f800b258a0ba40b5253077702fa6befce4b1b4c0 Mon Sep 17 00:00:00 2001
+From: Hongxu Jia <hongxu.jia@windriver.com>
+Date: Fri, 24 Dec 2021 10:07:52 +0800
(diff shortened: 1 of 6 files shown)


kraj:
too much of band-aid to support centos7 which is on the fag end of its support. I am not sure if its vaulable for future.

hongxu-jia:
> too much of band-aid to support centos7 which is on the fag end of its support. I am not sure if its valuable for future.
@kraj The CentOS7 is one case to resolve, I think the purpose of these commits is to support out of system-wide GCC, especially Yocto GCC provided by nativesdk-gcc in Yocto SDK. It is also helpful for nativesdk-clang to find Yocto GCC(nativesdk-gcc) rather than system GCC correctly

kraj:
> > too much of band-aid to support centos7 which is on the fag end of its support. I am not sure if its valuable for future.
> 
> @kraj The CentOS7 is one case to resolve, I think the purpose of these commits is to support out of system-wide GCC, especially Yocto GCC provided by nativesdk-gcc in Yocto SDK. It is also helpful for nativesdk-clang to find Yocto GCC(nativesdk-gcc) rather than system GCC correctly

centos7 will be soon moved to unsupported in master as yocto autobuilder has dropped it as well. Therefore I am less inclined to apply these patches now.

PR #550, opened 2022-01-05
intel-mediasdk: Fix meta-intel/meta-clang collisions

esben:
meta-intel assumes LLVM 10 or 11 is being built, and forces a version
of SPIRV-LLVM-Translator that doesn't work with LLVM 12.
Override this with an LLVM 12 compatible version.

Also, meta-intel pulls in intel-mediasdk which, out of the box, doesn't
build with compiler-rt. Add a patch to intel-mediasdk that fixes this.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/dynamic-layers/intel/recipes-multimedia/mediasdk/intel-mediasdk/intel-mediasdk-20.5.1-no-lgcc.patch
+++ b/dynamic-layers/intel/recipes-multimedia/mediasdk/intel-mediasdk/intel-mediasdk-20.5.1-no-lgcc.patch
@@ -1,0 +1,30 @@
+Author: Bernhard Rosenkraenzer <bernhard.rosenkraenzer.ext@huawei.com>
+
+Remove -lgcc -- it's unneeded and breaks building with compiler-rt.
+
+Signed-off-by: Bernhard Rosenkraenzer <bernhard.rosenkraenzer.ext@huawei.com>
+Upstream-Status: Pending
+
+diff -up git/builder/FindFunctions.cmake.omv~ git/builder/FindFunctions.cmake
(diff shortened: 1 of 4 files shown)


recipes-devtools/clang/llvm-project-source.bbappend
```diff
@@ -0,0 +1,9 @@
+# This needs to override meta-intel (which pulls in the
+# LLVM 10/11 version of SPIRV-LLVM-Translator)
+
+SPIRV_BRANCH = "llvm_release_120"
```
kraj, line 4: does this also work with clang-13?

dynamic-layers/intel/recipes-multimedia/mediasdk/intel-mediasdk/intel-mediasdk-20.5.1-no-lgcc.patch
```diff
@@ -0,0 +1,30 @@
...
+ 
+@@ -278,7 +274,7 @@ function( make_executable name variant )
+   endforeach()
+ 
+   if( Linux )
+-    target_link_libraries( ${target} "-Xlinker --end-group -lgcc" )
++    target_link_libraries( ${target} "-Xlinker --end-group" )
+   endif()
+ 
+   set( target ${target} PARENT_SCOPE )
```
kraj, line 30: this patch should be in meta-intel IMO, have you proposed it there ?

esben: Yes, this belongs in meta-intel or intel-mediasdk upstream.  It will be proposed there instead.
Closing.

PR #555, opened 2022-01-14
Enable CI for honister

rossburton:
I'm not sure if this is the right thing, but lets see what the CI does :)


rossburton:
Oh, no yoe for honister. 😭

PR #606, opened 2022-04-12
compiler-rt: Build builtins only for the default target

esben:
We were never really supposed to build for multiple target architectures in one
compiler-rt build, so let's use the feature created to do exactly that.

This also happens to fix building for ARM softfp ABI builds, where the target
architecture guessing logics ends up with:

    -- Builtin supported architectures: arm;armv6m

And later on bailing out because of duplicate ninja rules:

    ninja: error: build.ninja:56765: multiple rules generate lib/clang/14.0.0/lib/linux/libclang_rt.builtins-arm.a [-w dupbuild=err]

Signed-off-by: Esben Haabendal <esben.haabendal@huawei.com>

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):
lgtm

PR #607, opened 2022-04-12
Nativesdk

esben:
---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/compiler-rt_git.bb
+++ b/recipes-devtools/clang/compiler-rt_git.bb
@@ -22,8 +22,13 @@ INHIBIT_DEFAULT_DEPS = "1"
 
 DEPENDS += "ninja-native libgcc"
 DEPENDS:append:class-target = " clang-cross-${TARGET_ARCH} virtual/${MLPREFIX}libc gcc-runtime"
-DEPENDS:append:class-nativesdk = " clang-native"
+DEPENDS:append:class-nativesdk = " clang-native clang-crosssdk-${SDK_ARCH} nativesdk-gcc-runtime"
 DEPENDS:append:class-native = " clang-native"
+
+# Trick clang.bbclass into not creating circular dependencies
+UNWINDLIB:class-nativesdk = "--unwindlib=libgcc"
+COMPILER_RT:class-nativesdk:toolchain-clang:runtime-llvm = "-rtlib=libgcc --unwindlib=libgcc"
+LIBCPLUSPLUS:class-nativesdk:toolchain-clang:runtime-llvm = "-stdlib=libstdc++"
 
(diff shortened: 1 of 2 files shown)


recipes-devtools/clang/compiler-rt_git.bb
```diff
@@ -22,9 +22,14 @@ INHIBIT_DEFAULT_DEPS = "1"
...
 DEPENDS += "ninja-native libgcc"
 DEPENDS:append:class-target = " clang-cross-${TARGET_ARCH} virtual/${MLPREFIX}libc gcc-runtime"
-DEPENDS:append:class-nativesdk = " clang-native"
+DEPENDS:append:class-nativesdk = " clang-native clang-crosssdk-${SDK_ARCH} nativesdk-gcc-runtime"
 DEPENDS:append:class-native = " clang-native"
 
+# Trick clang.bbclass into not creating circular dependencies
+UNWINDLIB:class-nativesdk = "--unwindlib=libgcc"
+COMPILER_RT:class-nativesdk:toolchain-clang:runtime-llvm = "-rtlib=libgcc --unwindlib=libgcc"
+LIBCPLUSPLUS:class-nativesdk:toolchain-clang:runtime-llvm = "-stdlib=libstdc++"
```
kraj, line 31: perhaps we can use `LDFLAGS:nativesdk` and `CXXFLAGS:nativesdk` here instead ?

kraj (commented):
lgtm, although maybe using `LDFLAGS:nativesdk` and `CXXFLAGS:nativesdk`

recipes-devtools/clang/compiler-rt_git.bb
esben: Ok, so on assumption that compiler-rt is only going to be built if `toolchain-clang` and `runtime-llvm` is true anyway?

kraj: right

esben: And you are sure it will not cause problems for those using `RUNTIME = "gnu"` ?

And you mean `LDFLAGS:class-nativesdk` and `CXXFLAGS:class-nativesdk` right?

kraj: yes thats correct. since we are forcing it to link with GCC runtime we are virtually saying RUNTIME=gnu anyway so it should work ok in that case too. It will also be logically similar to native case.

esben: Ok, I have strippted it down to `:class-nativesdk` and has verified that ti WORKSFORME.

PR #613, opened 2022-05-19
bcc: De-vendor libbpf

daniel-thompson:
Currently bcc builds against a vendored copy of libbpf. This causes problems for bpftrace which is built against bcc and the system libbpf. The resulting version skew between the vendored and system versions of libbpf resulting in a SEGV whenever bpftrace is used.

Although `--help` and `-l` (list probe points) work OK that is because they do not actually use BPF! Anything that does use BPF will crash immediately,including `bpftrace --info` and simple one-liners such as:

    ~# bpftrace -e 't:raw_syscalls:sys_exit { printf("%s", comm); }'
    Attaching 1 probe...
    Segmentation fault

Fixes: https://github.com/iovisor/bpftrace/issues/2173
Fixes: bb3e56b06f9d ("bpftrace: Fix build with new libbpf")
    
Currently I have fixed this in two parts, firstly we make de-vendoring work by using a nasty hack and secondly we make the actual change!

Second part is (IMHO) "good" but I am concerned the first part is better handled elsewhere. The data structures I have introduced are copyed from v5.17 kernel headers and I not entirely clear whether or not they are candidates to backport into linux-libc-headers in kirkstone. In other words I wanted to document current status and ask for advice on how to land the needed changes (and I can move that to the mailing lists if needed now that I have something to point at ;-) ). If nothing else I hope the patch does, at least, clearly illustrate the problem.

### Contributor checklist

  + Fully tested on kirkstone and (AFAICT) all contributing components (linux-libc-headers, libbpf, bcc, bpftrace) have not changed in master.

### Reviewer Guidelines

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-devtools/bcc/bcc/0001-Vendor-just-enough-extra-headers-to-allow-libbpf-to-.patch
+++ b/dynamic-layers/openembedded-layer/recipes-devtools/bcc/bcc/0001-Vendor-just-enough-extra-headers-to-allow-libbpf-to-.patch
@@ -1,0 +1,124 @@
+From 6cffc195eca6b53a12865d325ff97e7b5ba8f22b Mon Sep 17 00:00:00 2001
+From: Daniel Thompson <daniel.thompson@linaro.org>
+Date: Thu, 19 May 2022 09:14:20 +0000
+Subject: [PATCH] Vendor just enough extra headers to allow libbpf to be
+ de-vendored
+
+Currently it is not possible to build the bcc recipe when we de-vendor
+libbpf and adopt the packaged version. Ironically this is due to the
+deliberate vendoring of some depreciated btf code that is being removed
+upstream because bcc was the only user! In other words the vendored code
+doesn't work the de-vendored libbpf because also ends up de-vendoring
+one of the Linux uapi headers.
+
+This is obviously an OE specific issue (due to the current combination
+of linux headers, libbpf and bcc). It's a bit of a mess and the right
+solution is probably to update the system UAPI headers but I am committing
+this for now simply so I can easily show why we must de-vendor libbpf in
+the first place!
+
+Upstream-status: Inappropriate [other]
+
+Signed-off-by: Daniel Thompson <daniel.thompson@linaro.org>
+---
+ src/cc/bcc_btf.cc | 87 +++++++++++++++++++++++++++++++++++++++++++++++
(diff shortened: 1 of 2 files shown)


dynamic-layers/openembedded-layer/recipes-devtools/bcc/bcc/0001-Vendor-just-enough-extra-headers-to-allow-libbpf-to-.patch
```diff
@@ -0,0 +1,124 @@
...
+deliberate vendoring of some depreciated btf code that is being removed
+upstream because bcc was the only user! In other words the vendored code
+doesn't work the de-vendored libbpf because also ends up de-vendoring
+one of the Linux uapi headers.
+
+This is obviously an OE specific issue (due to the current combination
+of linux headers, libbpf and bcc). It's a bit of a mess and the right
+solution is probably to update the system UAPI headers but I am committing
+this for now simply so I can easily show why we must de-vendor libbpf in
+the first place!
```
kraj, line 18: this patch has significant cost of maintenance in future if it  remains here, so I am a bit worried. Have you discussed this with upstream bcc community to see how they want to handle this situation for future ?

kraj: I see, you mean linux UAPI headers right ? which we maybe doing soon,

daniel-thompson: Yep, UAPI headers. If that gets updated then it should be easy to get bpftrace working on master by de-vendoring the bcc build.

I assume kirkstone will stay on the older UAPI though, right?

BTW and FYI, upstream discussion is here: https://github.com/iovisor/bcc/issues/4014

kraj: Yes Kirk stone will stay where it is

b49020: Hi @kraj,

Can you help backport this patch to kirkstone stable branch?

-Sumit

PR #633, opened 2022-07-27
clang: Build float128 soft float builtins for x86_64

quaresmajose:
Signed-off-by: Jose Quaresma <jose.quaresma@foundries.io>

---
### Contributor checklist

### Reviewer Guidelines


quaresmajose:
One side effect of this patch is that it breaks the **mpfr** build with clang https://github.com/foundriesio/meta-lmp/commit/8154f36a67c9c9357d0ebc093c31f6efd362b0b6

kraj:
@quaresmajose thanks for the patch, I see that https://reviews.llvm.org/D53608 still has some changes planned, so its not yet final, so we should wait for taking this patch into release. It might need to address the feedback and go through some more review cycles.

quaresmajose:
Thanks for the review, I will keep an eye on it and maybe can try to address the feedback.

quaresmajose:
there is no nativity upstream on https://reviews.llvm.org/D53608 and so close this.

PR #678, opened 2022-10-13
Fix compiler-rt-sanitizers to be able to build on arm targets

fwuehr95:
Signed off: Florian Wühr [fwuehr@rational-online.com](mailto:fwuehr@rational-online.com)
Signed off: Alexander Thoma [athoma@rational-online.com](mailto:athoma@rational-online.com)
---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/compiler-rt-sanitizers_git.bb
+++ b/recipes-devtools/clang/compiler-rt-sanitizers_git.bb
@@ -15,11 +15,17 @@ LIC_FILES_CHKSUM = "file://compiler-rt/LICENSE.TXT;md5=d846d1d65baf322d4c485d6ee
 
 TUNE_CCARGS_remove = "-no-integrated-as"
 
-RUNTIME = "llvm"
+#compiler-rt-sanitizers will not compile for arm32 or arm64 based platforms without adding -fPIC
+CXXFLAGS:append = " -fPIC"
+
+#set RUNTIME variable according to documention to gnu as default value
+RUNTIME ?= "gnu"
 
 DEPENDS += "ninja-native clang-cross-${TARGET_ARCH} virtual/${MLPREFIX}libc virtual/${TARGET_PREFIX}compilerlibs"
 DEPENDS_append_libc-glibc = " libxcrypt"
 DEPENDS_append_class-nativesdk = " clang-native nativesdk-libxcrypt"
+
+
 
 PACKAGECONFIG ??= ""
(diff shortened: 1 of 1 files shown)


fwuehr95:
We noticed that for arm32 or arm64 based targets it is not possible to build compiler-rt-sanitizers without adding "-fPIC" to CXXFLAGS. For x86-64 based targets it looks like it doesn´t really matter. Introducing "-fPIC" leads to package_qa errors with relocations in .text sections, which we have disable by adding "textrel" to INSANE_SKIP. This looks to be a known issue, other layer maintainers such as on meta-virtualization and openembedded-core used the same way here.

Additionally we changed the default value of RUNTIME variable, which has been set hardcoded to "llvm" prior to our change. According to meta-clang layer documentation RUNTIME should be set to "gnu" as default. Assigning "gnu" only with "?=" instead of "=" assignment allows to set RUNTIME variable to the desired value in a central conf file such as distro or machine configuration and to be used as a global variable. 

This helped us fixing the issue that we are not able of compiling libcxx for our arm32 and arm64 targets and to use GNU libstdc++ instead.

recipes-devtools/clang/compiler-rt-sanitizers_git.bb
```diff
@@ -15,12 +15,18 @@ LIC_FILES_CHKSUM = "file://compiler-rt/LICENSE.TXT;md5=d846d1d65baf322d4c485d6ee
 
 TUNE_CCARGS_remove = "-no-integrated-as"
 
-RUNTIME = "llvm"
+#compiler-rt-sanitizers will not compile for arm32 or arm64 based platforms without adding -fPIC
+CXXFLAGS:append = " -fPIC"
+
+#set RUNTIME variable according to documention to gnu as default value
+RUNTIME ?= "gnu"
```
kraj, line 22: gnu is default runtime, so we don't need to repeat it here.

kraj (changes requested):

recipes-devtools/clang/compiler-rt-sanitizers_git.bb
fwuehr95: Fine. Why has it been set to "llvm" previous to my changes though ?

kraj: this recipe is derived from compiler-rt which obviously is part of llvm runtime but sanitizers could be decoupled from runtime, so I am ok removing it.

PR #709, opened 2022-12-19
Fix OpenMP builds with thin LTO enabled.

jaskij:
When built with thin LTO enabled, LLVM produces object files containing LLVM IR, which ld can't understand. Since we're forcing the toolchain to clang anyway, let's also force using lld.

I was supposed to make this patch against master, but I can not reproduce it there, and my machine is too weak to repeatedly rebuild various clang versions in a reasonable time.

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):
lgtm

PR #719, opened 2023-01-09
openmp: skip buildpaths QA check for openmp-staticdev

rossburton:
The /usr/lib/libomptarget.devicertl.a static library ends up with build paths embedded.  The CMake files don't respect our CXXFLAGS so don't use the -fdebug-path-map arguments, but even adding those didn't solve it: it looks like code via goes via LLVM bytecode files doesn't get remapped?

Until that is solved, skip the buildpath check.

Signed-off-by: Ross Burton <ross.burton@arm.com>


kraj (approved):
lgtm

PR #729, opened 2023-01-27
Fix Upstream-Status formatting

shr-project:


kraj (approved):
lgtm, thanks for cleanup

PR #756, opened 2023-03-08
Add adb daemon

Ecordonnier:
---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/dynamic-layers/selinux/android-tools/android-tools/adbd.mk
+++ b/dynamic-layers/selinux/android-tools/android-tools/adbd.mk
@@ -1,0 +1,72 @@
+include ../../rules_yocto.mk
+NAME = adbd
+
+SOURCES = \
+	adb/daemon/main.cpp \
+	adb/daemon/auth.cpp \
+	adb/daemon/jdwp_service.cpp \
+	adb/daemon/file_sync_service.cpp \
+	adb/daemon/services.cpp \
+	adb/daemon/shell_service.cpp \
+	adb/daemon/remount_service.cpp \
+	adb/daemon/restart_service.cpp \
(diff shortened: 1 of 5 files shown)


kraj (commented):
whats the plan to upstream these patches to android

Ecordonnier:
Do you mean upstream to Debian? On Android the patches for adbd don't make sense. The makefiles don't exist on Android (they were introduced by Debian). I will try to upstream the makefiles patches to Debian.

kraj:
> Do you mean upstream to Debian? On Android the patches for adbd don't make sense. The makefiles don't exist on Android (they were introduced by Debian). I will try to upstream the makefiles patches to Debian.

debian, so we can share some of the maintenance

Ecordonnier:
I'll try to upstream what is upstreamable to Debian.

Ecordonnier:
@kraj thank you for merging this PR. I'm not sure what the best channel to discuss moving this recipe to meta-openembedded. Should I open an issue in github? Or maybe the meta-openembedded mailing list?

It is quite easily possible to compile this recipe without clang (just requires one straightforward patch), so this could be moved to meta-openembedded and make it usable without the clang dependency which is quite big in terms of compilation time. However it's not clear to me why the version from meta-openembedded compiles its own libselinux instead of depending on the generic libselinux. Is it done that way so that it doesn't need to be in meta-oe/dynamic-layers/selinux?

Also, this recipe, while newer than the version in meta-openembedded, may break some functionality provided by the patches in meta-openembedded, so there is a risk of regression.

Do you agree it makes sense to move it to meta-openembedded/meta-oe/dynamic-layers/selinux ?

PR #757, opened 2023-03-09
Layer cleanups (part 1)

rossburton:
Some small layer cleanups to start with...

### DIFF
--- a/recipes-graphics/mesa/mesa_%.bbappend
+++ b/recipes-graphics/mesa/mesa_%.bbappend
@@ -1,6 +1,3 @@
-LDFLAGS:append:toolchain-clang = " -latomic -lm"
-DEPENDS:append:toolchain-clang = " libatomic-ops"
-
 EXTRA_OEMASON:append:toolchain-clang:x86 = " -Dasm=false"
 EXTRA_OEMASON:append:toolchain-clang:x86-64 = " -Dasm=false"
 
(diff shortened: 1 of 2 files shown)


recipes-graphics/mesa/mesa_%.bbappend
```diff
@@ -1,6 +1,3 @@
-LDFLAGS:append:toolchain-clang = " -latomic -lm"
-DEPENDS:append:toolchain-clang = " libatomic-ops"
```
kraj, line 4: I think the real test is mips and riscv32 or ppc32 where 64bit builtin atomics are missing

recipes-graphics/mesa/mesa_%.bbappend
```diff
@@ -1,6 +1,3 @@
-EXTRA_OEMASON:append:toolchain-clang:x86 = " -Dasm=false"
-EXTRA_OEMASON:append:toolchain-clang:x86-64 = " -Dasm=false"
```
kraj, line 7: hmmm interesting find.

recipes-graphics/mesa/mesa_%.bbappend
rossburton: `MACHINE=qemumips bitbake mesa-demos` just build successfully. Is that enough?

kraj: seems good thanks

PR #761, opened 2023-03-15
android-tools: remove recipe and dynamic-layer

Ecordonnier:
The recipe has been patched to remove the clang dependency and moved to meta-openembedded at https://github.com/openembedded/meta-openembedded/tree/master/meta-oe/dynamic-layers/selinux/recipes-devtool/android-tools

---
### Contributor checklist

### Reviewer Guidelines


Ecordonnier:
There seems to be some issue with the test setup. I don't think the test failure is caused by the branch.

PR #766, opened 2023-03-20
layer.conf: use CLANGLLVMVERSION instead of LLVMVERSION

shr-project:
* the case where this seems to work incorrectly is meta-clang clang recipe which correctly says:
```
  conf/layer.conf:LLVMVERSION = "15.0.7"
  conf/layer.conf:PREFERRED_PROVIDER_llvm-native = "clang-native"
  recipes-devtools/clang/clang_git.bb:PROVIDES:append:class-native = " llvm-native"
```

  LLVMVERSION is used by oe-core as:

```
meta/conf/distro/include/tcmode-default.inc:LLVMVERSION ?= "15.%"
meta/conf/distro/include/tcmode-default.inc:PREFERRED_VERSION_llvm = "${LLVMVERSION}"
meta/conf/distro/include/tcmode-default.inc:PREFERRED_VERSION_llvm-native = "${LLVMVERSION}"
meta/conf/distro/include/tcmode-default.inc:PREFERRED_VERSION_nativesdk-llvm = ${LLVMVERSION}"
```

  and now this code will cause a warning about 15.0.7 version not being available even when in the end it will use clang-native as a provider with this version:
```
  $ bitbake-getvar -r llvm-native PV WARNING: preferred version 15.0.7 of llvm-native not available (for item llvm-native) WARNING: versions of llvm-native available: 15.0.6 # # $PV [4 operations]
  #   set oe-core/meta/conf/bitbake.conf:239
  #     "${@bb.parse.vars_from_file(d.getVar('FILE', False),d)[1] or '1.0'}"
  #   set oe-core/meta/conf/documentation.conf:340
  #     [doc] "The version of the recipe. The version is normally extracted from the recipe filename."
  #   set oe-core/meta/classes-global/sstate.bbclass:51
  #     [vardepvalue] "${PV}"
  #   set meta-clang/recipes-devtools/clang/clang.inc:13
  #     "${MAJOR_VER}.${MINOR_VER}.${PATCH_VER}"
  # pre-expansion value:
  #   "${MAJOR_VER}.${MINOR_VER}.${PATCH_VER}"
  PV="15.0.7"
```
  the problem is that both llvm-native from oe-core and clang-native are found as possible providers for item llvm-native, but then we call findPreferredProvider for pkg_pn clang-native, we'll check for PREFERRED_VERSION_clang_native not PREFERRED_VERSION_llvm-native

  then when running findPreferredProvider for pkg_pn llvm-native we find the PREFERRED_VERSION_llvm-native and if it doesn't match with PV we issue the warning, be aware that this issue is reproducible only when they don't happen to match, e.g. with master before: https://git.openembedded.org/openembedded-core/commit/?id=7438f9a9efaf0826f1be13acbc51f0a1134cf175 in dunfell it's triggered as well, with different versions:
```
  WARNING: preferred version 14.0.3 of llvm-native not available (for item llvm-native)
  WARNING: versions of llvm-native available: 13.0.1
```

* if we don't set P_V_llvm-native to match clang-native version, we can still depend on P_P_llvm-native to select clang as a provider and leave llvm-native to match the version of llvm recipe in oe-core


kraj (commented):
Thanks for trying to fix it. It has been coming up on and off other times too. There are other layers which encode `LLVMVERSION` as well. E.g. see

```
poky/meta/conf/distro/include/tcmode-default.inc:LLVMVERSION ?= "15.%"
poky/meta/conf/distro/include/tcmode-default.inc:PREFERRED_VERSION_llvm = "${LLVMVERSION}"
poky/meta/conf/distro/include/tcmode-default.inc:PREFERRED_VERSION_llvm-native = "${LLVMVERSION}"
poky/meta/conf/distro/include/tcmode-default.inc:PREFERRED_VERSION_nativesdk-llvm = "${LLVMVERSION}"
meta-intel/dynamic-layers/clang-layer/recipes-opencl/igc/intel-graphics-compiler_1.0.12812.24.bb:                  -DIGC_OPTION__LLVM_PREFERRED_VERSION=${LLVMVERSION} \
meta-intel/dynamic-layers/clang-layer/recipes-opencl/opencl-clang/opencl-clang_14.0.0.bb:                  -DPREFERRED_LLVM_VERSION=${LLVMVERSION} \
meta-intel/dynamic-layers/clang-layer/recipes-opencl/opencl-clang/opencl-clang_15.0.0.bb:                  -DPREFERRED_LLVM_VERSION=${LLVMVERSION} \
meta-intel/conf/machine/include/meta-intel.inc:PREFERRED_VERSION_opencl-clang ?= "${@bb.utils.contains('LLVMVERSION', '14.0.3', '14.0.0', '15.0.0', d)}"
meta-intel/conf/machine/include/meta-intel.inc:PREFERRED_VERSION_opencl-clang-native ?= "${@bb.utils.contains('LLVMVERSION', '14.0.3', '14.0.0', '15.0.0', d)}"
meta-riscv/recipes-graphics/mesa/mesa-pvr.inc:MESA_LLVM_RELEASE ?= "${LLVMVERSION}"
meta-browser/meta-chromium/recipes-browser/chromium/chromium-gn.inc:# Check the LLVMVERSION defined in the meta-clang layer. Given Chromium is
meta-browser/meta-chromium/recipes-browser/chromium/chromium-gn.inc:# developed using new C++ features, the LLVMVERSION has to be >= 12. Otherwise,
meta-browser/meta-chromium/recipes-browser/chromium/chromium-gn.inc:  llvm_version = d.getVar('LLVMVERSION', False)
meta-browser/meta-chromium/recipes-browser/chromium/chromium-gn.inc:    bb.fatal("Your LLVMVERSION (%s) is lower than the minimum required "
meta-browser/meta-chromium/recipes-browser/chromium/chromium-gn.inc:             "LLVMVERSION (%s). If you are using dunfell, make sure you "
meta-clang/conf/layer.conf:LLVMVERSION = "16.0.0"
meta-clang/recipes-devtools/spirv-llvm-translator/spirv-llvm-translator_git.bb:        -DBASE_LLVM_VERSION=${LLVMVERSION} \
meta-clang/dynamic-layers/openembedded-layer/recipes-devtools/bcc/bcc_0.26.0.bb:    -DLLVM_PACKAGE_VERSION=${LLVMVERSION} \
meta-clang/dynamic-layers/openembedded-layer/recipes-devtools/bpftrace/bpftrace_0.17.0.bb:    pvsplit = d.getVar('LLVMVERSION').split('.')
```

How will it function in such scene?

shr-project:
> How will it function in such scene?

You're right, badly, it doesn't even work right with older oe-core where LLVMVERSION was used for more things than just P_V_*llvm* e.g. in kirkstone and langdale as reported in:
https://lists.openembedded.org/g/bitbake-devel/message/14523

And when testing with mickledore it worked for me with oe-core and few other layers I regularly use, but I didn't use meta-intel/meta-risc/meta-browser with it and you're right it would probably cause issues there.

Maybe we should use different variable name for P_V_*llvm* use it just for that and leave LLVMVERSION for other uses where it doesn't have to match with llvm recipe version.

Or just continue to live with warnings. It's just confusing as bitbake says it's not available when in the end it uses the right preferred provider with expected version and since I'm using separate log file for warnings now I wanted to make my builds warning-free even more :).

I've tried to figure our some better solution as documented in:
https://lists.openembedded.org/g/bitbake-devel/message/14521
thread.

> maybe we should just BBMASK llvm-native from core, since I would like to use the version from meta-clang when its in mix.

I don't see how that would work, because the right llvm *is* being used, it's just that bitbake warns that P_V_*llvm* in oe-core is set to unavailable version (even when we don't end building llvm recipes because clang is preferred provider).

rossburton:
Our CI has fatal warnings and we just hit this issue.

shr-project:
Another alternative solution sent to oe-core:
https://lists.openembedded.org/g/openembedded-core/message/179469

kraj:
> Another alternative solution sent to oe-core: https://lists.openembedded.org/g/openembedded-core/message/179469

this one looks less painful.

PR #786, opened 2023-05-10
LLVM officially supports LoongArch starting from 16.0.0

baibaidashixiong:
Hi, LLVM officially supports LoongArch starting from 16.0.0, we can add its support to meta-clang now.


kraj (approved):
lgtm

PR #793, opened 2023-05-13
clang.bbclass: create lld symbolic link in sysroot when ld-is-lld is set

liuming50:
Create a symbolic link lld -> ld in recipe sysroot to choose the default linker lld when ld-is-lld is set in DISTRO_FEATURES. othereise, we can get linking issues when '-fuse-ld=lld' is in LDFLAGS but the actual ld is not lld.

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):
lgtm

PR #800, opened 2023-05-25
Update README.md

patrick-dojofive:
Updating to latest URIs used by GitHub.com

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,10 @@ with GNU compiler and can be used for specific recipes or full system compiler.
 # Getting Started
 
 ```shell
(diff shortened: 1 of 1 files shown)


patrick-dojofive:
> This switches to using ssh protocol, I agree that we should not mention anonymous git protocol but ssh protocol is only going to work for folks who have write access to the repo. So I would say switch to https instead which is preferred protocol and would be the right url for all users.

Well it's a public repo, so all _users_ can access it, but https allows anonymous access.

kraj (approved):
lgtm

kraj (approved):

PR #813, opened 2023-06-21
llvm-config: fix script for multilib build

saininav:
From multilib build:
| CMake Error at cmake/FindLLVM.cmake:79 (message):
|   Failed running
|  tmp-glibc/work/x86-64-v3-oe-linux/ispc/1.20.0-r0/recipe-sysroot/usr/bin/crossscripts/llvm-config;--libfiles;engine;ipo;bitreader;bitwriter;instrumentation;linker;option;frontendopenmp;windowsdriver;x86
| Call Stack (most recent call first):
|   cmake/FindLLVM.cmake:116 (run_llvm_config)
|   CMakeLists.txt:377 (get_llvm_libfiles)

Currently hardcoded to /lib, which causing multilib build failure.

An easy solution could be set and export YOCTO_ALTERNATE_LIBDIR to $baselib in recipe, but it would require to add this in all affected recipes.

Other approach is to handle in llvm-config script itself. Unfortunately $baselib is not available in env for the llvm-config script, so extracting baselib value from $base_libdir as its being exported and available in env and then set YOCTO_ALTERNATE_LIBDIR accordingly.

---
### Contributor checklist

### Reviewer Guidelines


recipes-devtools/clang/clang/llvm-config
```diff
@@ -8,7 +8,11 @@
 SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
 NEXT_LLVM_CONFIG="$(which -a llvm-config | sed -n 2p)"
 export YOCTO_ALTERNATE_EXE_PATH="${YOCTO_ALTERNATE_EXE_PATH:="$(readlink -f "$SCRIPT_DIR/../llvm-config")"}"
-export YOCTO_ALTERNATE_LIBDIR="${YOCTO_ALTERNATE_LIBDIR:="/lib"}"
+if [[ "$base_libdir" == *"lib64"* ]]; then
```
kraj, line 11: `base_libdir` with `usrmerge` distro feature will prepend `/usr` to lib64, not that this patch is doing it but this will be a problem. Secondly, Using bashism needs to be removed as well.

saininav: Yes, base_libdir consider usrmerge for prefix, but it also read value from baselib. That's why I am checking only for lib64 substring in base_libdir value. If I understood correctly !

From bitbake.conf:

root_prefix = "${@bb.utils.contains('DISTRO_FEATURES', 'usrmerge', '${exec_prefix}', '${base_prefix}', d)}"
export base_libdir = "${root_prefix}/${baselib}"

kraj (approved):
lgtm

recipes-devtools/clang/clang/llvm-config
kraj: btw. dont we need it in master too ? if so please post a pull for master as well.

saininav:
V2: Removed bashism [[ ]] and *""*

PR #815, opened 2023-06-22
llvm-config: fix script for multilib build

saininav:
From multilib build:
| CMake Error at cmake/FindLLVM.cmake:79 (message):
|   Failed running
|  tmp-glibc/work/x86-64-v3-oe-linux/ispc/1.20.0-r0/recipe-sysroot/usr/bin/crossscripts/llvm-config;--libfiles;engine;ipo;bitreader;bitwriter;instrumentation;linker;option;frontendopenmp;windowsdriver;x86
| Call Stack (most recent call first):
|   cmake/FindLLVM.cmake:116 (run_llvm_config)
|   CMakeLists.txt:377 (get_llvm_libfiles)

Currently hardcoded to /lib, which causing multilib build failure.

An easy solution could be set and export YOCTO_ALTERNATE_LIBDIR to $baselib in recipe, but it would require to add this in all affected recipes.

Other approach is to handle in llvm-config script itself. Unfortunately $baselib is not available in env for the llvm-config script, so extracting baselib value from $base_libdir as its being exported and available in env and then set YOCTO_ALTERNATE_LIBDIR accordingly.

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):
lgtm

PR #832, opened 2023-08-03
Fix elfutils and gnu-efi build with LLD linker

kraj:
---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/.github/workflows/yoe.yml
+++ b/.github/workflows/yoe.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Build Image
         run: |
           cd yoe
-          /bin/bash -c ". ./envsetup.sh qemuarm64 && bitbake yoe-qt5-image yoe-sdk-image yoe-debug-image"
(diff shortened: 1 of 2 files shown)


classes/clang.bbclass
```diff
@@ -66,14 +67,18 @@ LDFLAGS:toolchain-clang:class-nativesdk = "${BUILDSDK_LDFLAGS} \
...
+LDFLAGS:append:toolchain-clang = " -fuse-ld=${LINKER}"
 
 # Remove gcc specific -fcanon-prefix-map option, added in gcc-13+
 # clang does not support it yet
 DEBUG_PREFIX_MAP:remove:toolchain-clang = "-fcanon-prefix-map"
 
 # choose between 'gcc' 'clang' an empty '' can be used as well
 TOOLCHAIN ??= "gcc"
+# choose one of bfd, gold, lld ( may be 'mold' too )
+LINKER ??= "bfd"
```
quaresmajose, line 81: I think this has no effect because LINKER is previous assigned on line 71

kraj: right I have taken care of it in v2. Thanks for reviewing.

PR #841, opened 2023-08-18
clang.bbclass: fix recipe_sysroot_check_ld_is_lld

jaskij:
Currently, the check added in #802 breaks my builds. I have been able to find, and fix, two cases:

- allarch recipes, where lld is simply not installed in native sysroot
- for some recipes (`ed` for example), `${bindir}` evaluates to `/bin`, not `/usr/bin`, which results in incorrect path to lld

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/classes/clang.bbclass
+++ b/classes/clang.bbclass
@@ -142,7 +142,9 @@ RECIPESYSROOTFUNCS = ""
 RECIPESYSROOTFUNCS:toolchain-clang = "recipe_sysroot_check_ld_is_lld"
 
 recipe_sysroot_check_ld_is_lld () {
-    if "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}"; then
+    if "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}" && \
+        "${@bb.utils.contains('TARGET_ARCH', 'allarch', 'false', 'true', d)}"
+    then
         ln -srf ${RECIPE_SYSROOT_NATIVE}${bindir}/${TARGET_SYS}/${TARGET_PREFIX}ld.lld ${RECIPE_SYSROOT_NATIVE}${bindir}/${TARGET_SYS}/${TARGET_PREFIX}ld
     fi
 }
--- a/classes/clang.bbclass
+++ b/classes/clang.bbclass
@@ -145,7 +145,7 @@ recipe_sysroot_check_ld_is_lld () {
     if "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}" && \
         "${@bb.utils.contains('TARGET_ARCH', 'allarch', 'false', 'true', d)}"
     then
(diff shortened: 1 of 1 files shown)


jaskij:
While those two changes fix most of my build errors after pulling in latest Kirkstone across the board, there is one more package I have failing (`depmodwrapper-cross`) which also complains about the `ln` call in this task.

jaskij:
cc @liuming50 , as the author of the original patch

jaskij:
Ah, `depmodwrapper-cross` is a special case, it's a script, but probably not truly suited to being marked `allarch`. Still, I did that as a workaround.

Hmm... perhaps just test if the file exists?

```bash
[ -f "${STAGING_BINDIR_NATIVE}/${TARGET_SYS}/${TARGET_PREFIX}ld.lld" ]
```

Anyway, removing draft designation.

classes/clang.bbclass
```diff
@@ -145,7 +145,7 @@ recipe_sysroot_check_ld_is_lld () {
     if "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}" && \
         "${@bb.utils.contains('TARGET_ARCH', 'allarch', 'false', 'true', d)}"
     then
-        ln -srf ${RECIPE_SYSROOT_NATIVE}${bindir}/${TARGET_SYS}/${TARGET_PREFIX}ld.lld ${RECIPE_SYSROOT_NATIVE}${bindir}/${TARGET_SYS}/${TARGET_PREFIX}ld
+        ln -srf ${STAGING_BINDIR_NATIVE}/${TARGET_SYS}/${TARGET_PREFIX}ld.lld ${STAGING_BINDIR_NATIVE}/${TARGET_SYS}/${TARGET_PREFIX}ld
```
kraj, line 148: Please backport 6ecaea06d5e6e7ed0a39b4c019fc485f18e1ce55 and see if that works

classes/clang.bbclass
```diff
@@ -142,7 +142,9 @@ RECIPESYSROOTFUNCS = ""
 RECIPESYSROOTFUNCS:toolchain-clang = "recipe_sysroot_check_ld_is_lld"
 
 recipe_sysroot_check_ld_is_lld () {
-    if "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}"; then
+    if "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}" && \
+        "${@bb.utils.contains('TARGET_ARCH', 'allarch', 'false', 'true', d)}"
+    then
```
kraj, line 147: you also need to backport 7419c1f4e13187311870785a11e8cffee2c640e6

classes/clang.bbclass
```diff
@@ -142,7 +142,9 @@ RECIPESYSROOTFUNCS = ""
 RECIPESYSROOTFUNCS:toolchain-clang = "recipe_sysroot_check_ld_is_lld"
 
 recipe_sysroot_check_ld_is_lld () {
-    if "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}"; then
+    if "${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)}" && \
+        "${@bb.utils.contains('TARGET_ARCH', 'allarch', 'false', 'true', d)}"
```
kraj, line 146: this is already fixed in master please do a direct backport of 0f31c0b29580825af8700e484b984a8824a31a4f

kraj (changes requested):

PR #875, opened 2023-10-29
clang.bbclass: fix a typo

liuming50:
A pair of '[]' is missing, add them in.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/classes/clang.bbclass
+++ b/classes/clang.bbclass
@@ -142,7 +142,7 @@ RECIPESYSROOTFUNCS:toolchain-clang = "recipe_sysroot_check_ld_is_lld"
 
 recipe_sysroot_check_ld_is_lld () {
     if ${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)} &&  \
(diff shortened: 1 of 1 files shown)


classes/clang.bbclass
```diff
@@ -142,7 +142,7 @@ RECIPESYSROOTFUNCS:toolchain-clang = "recipe_sysroot_check_ld_is_lld"
 
 recipe_sysroot_check_ld_is_lld () {
     if ${@bb.utils.contains('DISTRO_FEATURES', 'ld-is-lld', 'true', 'false', d)} &&  \
-        -e ${STAGING_BINDIR_TOOLCHAIN}/${TARGET_PREFIX}ld.lld; then
+        [ -e ${STAGING_BINDIR_TOOLCHAIN}/${TARGET_PREFIX}ld.lld ]; then
```
kraj, line 145: which shell shows the problem and what issue do you see ? can you post a simple shell example ?

liuming50: Hi, @kraj:

I was using bash, but this missing "[]" or "test" could lead to errors in any shell?

```
if true && -e /path/to/a/file; then echo; fi
```

result:
```
bash: -e: command not found
```

kraj (approved):

Issue #879, opened 2023-11-09
Kirkstone: libclang package is empty


fwuehr95:
Looking at the clang recipe in kirkstone branch FILES for the libclang package is set to the following:

FILES:libclang = "\
  ${libdir}/libclang.so.${MAJOR_VER} \
"
Unfortunately the llvm project has not changed the libclang version with the upgrade from 13 to 14, so there is a mismatch of the version naming of the resulting libclang (libclang.so.13) and the variable that shall be packaged which would equal to "libclang.so.14" (see: https://github.com/llvm/llvm-project/issues/53684).
As far as I understand it, they didn´t change the version due to the fact that there were no ABI changes. 

IMHO there are three options to resolve this: 

1. hardcode FILES:libclang to "${libdir}/libclang.so.13"
2. add an additional variable additionally to "MAJOR_VER" such as like "ABI_VERSION" if there is no steady relationship between these two numbers anymore
3. apply a patch to llvm-project-source at the file "CMakeLists.txt" of the libclang module and change the variable "CLANG_SONAME" to "14" instead of "13"

kraj:
I think it will be fine to hardcode it for 14 recipe, we do not allow multiple clang installs so it should be fine. I think with 15 onwards these versions match MAJOR_VER so we are good on master and newer releases.

fwuehr95:
Hi @kraj, thanks for the feedback, I opened a PR to fix this: https://github.com/kraj/meta-clang/pull/881

kraj:
Fixed with #881

PR #881, opened 2023-11-10
Fix packaging error for libclang on kirkstone

fwuehr95:
---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -307,7 +307,7 @@ FILES:${PN}-libllvm =+ "\
 "
 
 FILES:libclang = "\
-  ${libdir}/libclang.so.${MAJOR_VER} \
+  ${libdir}/libclang.so.13 \
 "
 
 FILES:${PN}-dev += "\


kraj (approved):
LGTM

miwojcik:
Hi @fwuehr95, @kraj 

I found out that after this commit `bpftrace` is broken:
```
root@qemuarm64:~# bpftrace -l
bpftrace: error while loading shared libraries: libclang.so.13: cannot open shared object file: No such file or directory```
```
and after more investigation I discovered potential problems with this change.

Before this commit, libclang package wasn't created as there were no files to package and both libclang.so.14.0.6 and a libclang.so.13 symlink were implicitly included in the clang package:
```
$ dpkg -c ./tmp/deploy/ipk/cortexa57/clang_14.0.6-r0_cortexa57.ipk
...
lrwxrwxrwx root/root         0 2011-04-06 01:00 ./usr/lib/libclang.so.13 -> libclang.so.14.0.6
-rwxr-xr-x root/root  29451416 2011-04-06 01:00 ./usr/lib/libclang.so.14.0.6
...
```
However, now it only contains the `libclang.so.13` symlink to the so file:
```
dpkg -c ./tmp/deploy/ipk/cortexa57/libclang_14.0.6-r0_cortexa57.ipk
drwxr-xr-x root/root         0 2011-04-06 01:00 ./usr/
drwxr-xr-x root/root         0 2011-04-06 01:00 ./usr/lib/
lrwxrwxrwx root/root         0 2011-04-06 01:00 ./usr/lib/libclang.so.13 -> libclang.so.14.0.6
```
and the so file is still in the clang package:
```
dpkg -c ./tmp/deploy/ipk/cortexa57/clang_14.0.6-r0_cortexa57.ipk
...
-rwxr-xr-x root/root  29451416 2011-04-06 01:00 ./usr/lib/libclang.so.14.0.6
...
```
The obvious fix would be to also include the so file in the libclang package, but here is my concern - we can't tell how many packages/recipes rely on the fact that libclang was implicitly included in the clang package, that would now require fixing the dependencies.
That would be a great fix on a master branch (although you mentioned in https://github.com/kraj/meta-clang/issues/879 it is not necessary), but kirkstone branch is LTS, so I'm not sure if it wouldn't be better to revert this commit and live with the fact that libclang is not packaged correctly, rather than to start a chain of dependency fixes.

Of course if you have other solutions I'd be happy to discuss them.

Temporarily I was able to bing back bpftrace to life by modifying its recipe:
```
--- a/dynamic-layers/openembedded-layer/recipes-devtools/bpftrace/bpftrace_0.14.1.bb
+++ b/dynamic-layers/openembedded-layer/recipes-devtools/bpftrace/bpftrace_0.14.1.bb
@@ -14,7 +14,7 @@ DEPENDS += "bison-native \
             "

 PV .= "+git${SRCREV}"
-RDEPENDS:${PN} += "bash python3 xz"
+RDEPENDS:${PN} += "bash libclang python3 xz"

 SRC_URI = "git://github.com/iovisor/bpftrace;branch=master;protocol=https \
            file://0001-Detect-new-BTF-api-btf_dump__new-btf_dump__new_v0_6_.patch \
```

PR #886, opened 2023-11-22
cmake-native.bbclass: Add ubfs specific variable to ignore list

kraj:
These variables show up as dependencies of native recipes otherwise can cause clang-native to rebuild without a reason just when defined by a machine using ubifs.

---
### Contributor checklist

### Reviewer Guidelines


shr-project:
How these variables end in cmake signature? I don't see it here, does some BSP export them or something? I think then the BSP layer should be fixed.

kraj:
they dont, the real problem was from BSP exporting these symbols for no reason. I will send a revert

PR #891, opened 2023-12-18
bpftrace: Replace python with python3 in ptest

wentao-windriver:
"runtime:call" in ptest gets the following FAILED: | python: No such file or directory
Replace python with python3 in this ptest item.

---
### Contributor checklist

- [x ] Changes have been tested
- [ x] `Signed-off-by` is present
- [x ] The PR complies with the [Open Embedded Commit Patch Message Guidelines](http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines)

### Reviewer Guidelines

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-devtools/bpftrace/bpftrace/0001-replace-python-with-python3-in-the-test.patch
+++ b/dynamic-layers/openembedded-layer/recipes-devtools/bpftrace/bpftrace/0001-replace-python-with-python3-in-the-test.patch
@@ -1,0 +1,53 @@
+From f2a61d19d8432a220184c4eed7c39eb042d0dfff Mon Sep 17 00:00:00 2001
+From: Wentao Zhang <wentao.zhang@windriver.com>
+Date: Tue, 15 Aug 2023 11:18:36 +0800
+Subject: [PATCH] replace python with python3 in the test
+
+"runtime:call" in ptest gets the following FAILED:
+python: No such file or directory
+replace python with python3 in the test scripts.
+
+$export BPFTRACE_RUNTIME_TEST_EXECUTABLE=/usr/bin
+$cd /usr/lib/bpftrace/ptest/tests
+$python3 runtime/engine/main.py --filter="call.*"
+***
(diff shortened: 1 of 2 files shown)


dynamic-layers/openembedded-layer/recipes-devtools/bpftrace/bpftrace/0001-replace-python-with-python3-in-the-test.patch
```diff
@@ -0,0 +1,53 @@
...
+[  FAILED  ] call.strftime_microsecond_extension_rollover
+	Command: /usr/bin/bpftrace -e 'BEGIN { printf("%s - %s\n", strftime
+    ("1%f", 1000000123000), strftime("1%f", 0)); exit(); }' | tail -n
+    +2 | xargs -I{} python -c "print({})"
+	Unclean exit code: 127
+	Output: __BPFTRACE_NOTIFY_PROBES_ATTACHED\nxargs: python: No such
+    file or directory\n
+***
+
+Upstream-Status: Inappropriate [oe specific]
```
kraj, line 24: if this patch work on non-OE envs then propose it upstream as well Please.

kraj (approved):

dynamic-layers/openembedded-layer/recipes-devtools/bpftrace/bpftrace/0001-replace-python-with-python3-in-the-test.patch
kraj: do we need this fix on master as well ? if so then please propose it for master first.

wentao-windriver: for 1st comment:
https://packages.debian.org/bookworm/python-is-python3
Debian has a package link python to python3. So I this this patch is not the only way to solve this problem for all non-OE envs.

for 2nd comment:
Yes, I think we need this fix on master as well, but I havn't test it yet. I will try it now.

PR #897, opened 2023-12-22
llvm-config: update following the oe-core script

lumag:
The Mesa Clover meson.build script passes two options to llvm-config script, --libs --ldflags. The script from meta-clang passes control to the native llvm-config script, which unfortunately results in the native dynamic linker option leaking to the cross build. Fix that by adopting the approach from OE-core and filter known options before calling into the native llvm-config.

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):
lgtm

PR #904, opened 2024-01-21
bcc: drop python3-distutils runtime dependency

shr-project:
distutils package no longer exists after python 3.12 upgrade.

I haven't run ptest for this, but at least it will parse well now instead of:
ERROR: Nothing RPROVIDES 'python3-distutils' (but meta-clang/dynamic-layers/meta-python/recipes-devtools/bcc/bcc_0.28.0.bb RDEPENDS on or otherwise requires it)

### DIFF
--- a/dynamic-layers/meta-python/recipes-devtools/bcc/bcc_0.28.0.bb
+++ b/dynamic-layers/meta-python/recipes-devtools/bcc/bcc_0.28.0.bb
@@ -22,7 +22,7 @@ LUAJIT:powerpc64 = ""
 LUAJIT:riscv64 = ""
 
 RDEPENDS:${PN} += "bash python3 python3-core python3-setuptools xz"
-RDEPENDS:${PN}-ptest = "cmake python3 python3-distutils python3-netaddr python3-pyroute2"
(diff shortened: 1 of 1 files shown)


dynamic-layers/meta-python/recipes-devtools/bcc/bcc_0.28.0.bb
```diff
@@ -22,7 +22,7 @@ LUAJIT:powerpc64 = ""
 LUAJIT:riscv64 = ""
 
 RDEPENDS:${PN} += "bash python3 python3-core python3-setuptools xz"
-RDEPENDS:${PN}-ptest = "cmake python3 python3-distutils python3-netaddr python3-pyroute2"
+RDEPENDS:${PN}-ptest = "cmake python3 python3-netaddr python3-pyroute2"
```
kraj, line 25: Thanks for the patch I have a local change where I bump to 0.29.1 as well but ptests have failures so perhaps it will be good to first try it with existing rev and I have clang 18 as well which might be playing part too

kraj (approved):

PR #912, opened 2024-02-12
clang: Allow conditional compilation of clangd

lorenzph:
clangd should not be needed in target builds. Extend PACKAGECONFIG to
enable selective disabling and disable it by default for target builds.

Additionally, this enables compilation of `clang-native` on Debian "bullseye"
hosts where GCC 10 fails to compile clangd with an internal compilation
error.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -55,30 +55,35 @@ def get_clang_target_arch(bb, d):
 def get_clang_experimental_target_arch(bb, d):
     return get_clang_experimental_arch(bb, d, 'TARGET_ARCH')
 
-PACKAGECONFIG ??= "compiler-rt libcplusplus shared-libs lldb-wchar \
-                   ${@bb.utils.filter('DISTRO_FEATURES', 'thin-lto lto', d)} \
-                   ${@bb.utils.contains('TC_CXX_RUNTIME', 'llvm', 'compiler-rt libcplusplus unwindlib libomp', '', d)} \
-                   rtti eh libedit terminfo \
-                   "
-PACKAGECONFIG:class-native = "rtti eh libedit shared-libs ${@bb.utils.contains('TC_CXX_RUNTIME', 'llvm', 'compiler-rt libcplusplus unwindlib libomp', '', d)}"
-PACKAGECONFIG:class-nativesdk = "rtti eh libedit shared-libs ${@bb.utils.filter('DISTRO_FEATURES', 'thin-lto lto', d)} ${@bb.utils.contains('TC_CXX_RUNTIME', 'llvm', 'compiler-rt libcplusplus unwindlib libomp', '', d)}"
+PACKAGECONFIG_CLANG_COMMON = "eh libedit rtti shared-libs \
+                              ${@bb.utils.contains('TC_CXX_RUNTIME', 'llvm', 'compiler-rt libcplusplus libomp unwindlib', '', d)} \
+                              "
 
-PACKAGECONFIG[compiler-rt] = "-DCLANG_DEFAULT_RTLIB=compiler-rt,,"
-PACKAGECONFIG[libcplusplus] = "-DCLANG_DEFAULT_CXX_STDLIB=libc++,,"
-PACKAGECONFIG[unwindlib] = "-DCLANG_DEFAULT_UNWINDLIB=libunwind,-DCLANG_DEFAULT_UNWINDLIB=libgcc,,"
-PACKAGECONFIG[libomp] = "-DCLANG_DEFAULT_OPENMP_RUNTIME=libomp,,"
-PACKAGECONFIG[thin-lto] = "-DLLVM_ENABLE_LTO=Thin -DLLVM_BINUTILS_INCDIR=${STAGING_INCDIR},,binutils,"
-PACKAGECONFIG[lto] = "-DLLVM_ENABLE_LTO=Full -DLLVM_BINUTILS_INCDIR=${STAGING_INCDIR},,binutils,"
-PACKAGECONFIG[shared-libs] = "-DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_LINK_LLVM_DYLIB=ON,,,"
-PACKAGECONFIG[terminfo] = "-DLLVM_ENABLE_TERMINFO=ON -DCOMPILER_RT_TERMINFO_LIB=ON,-DLLVM_ENABLE_TERMINFO=OFF -DCOMPILER_RT_TERMINFO_LIB=OFF,ncurses,"
-PACKAGECONFIG[pfm] = "-DLLVM_ENABLE_LIBPFM=ON,-DLLVM_ENABLE_LIBPFM=OFF,libpfm,"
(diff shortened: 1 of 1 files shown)


recipes-devtools/clang/clang_git.bb
```diff
@@ -55,13 +55,13 @@ def get_clang_target_arch(bb, d):
 def get_clang_experimental_target_arch(bb, d):
     return get_clang_experimental_arch(bb, d, 'TARGET_ARCH')
 
-PACKAGECONFIG ??= "compiler-rt libcplusplus shared-libs lldb-wchar \
+PACKAGECONFIG ??= "compiler-rt libcplusplus shared-libs lldb-wchar clangd \
```
lorenzph, line 58: I'm not sure how sensible it is to enable `clangd` in the target build (and to also deploy it to sysroot during preprocessing) but I've added to remain compatible with the current behaviour.

kraj: I think disabling it for target by default makes sense perhaps, however, disabling it for native is going to be a bit questionable, since clang-native is used as part of clang-cross as well, so I think having language server during cross builds is a fair ask. 

Can you try your builds against clang-18 pull request and see if this fixes the g++ ICE if it does then perhaps we can add this option and disable it only for target.

lorenzph: It looks like the ICE is gone with Clang 18. I'll update the PR to also enable clangd by default in the `native` build (and disable for the target).

quaresmajose:
it looks like `native` is replaced by `target` on the commit message.
looking at the code the clangd is disabled for class-target builds and not for native.
```
clangd is not needed for -native builds. Extend PACKAGECONFIG to enable
selective disabling and disable it by default for native builds.
```

lorenzph:
> it looks like `target` is replaced by `native` on the commit message. looking at the code the clangd is disabled for class-target builds and not for native.
> 
> ```
> clangd is not needed for -native builds. Extend PACKAGECONFIG to enable
> selective disabling and disable it by default for native builds.
> ```

My bad... Fixed.

recipes-devtools/clang/clang_git.bb
lorenzph: Actuall I stand corrected and it is still triggering an ICE:
```
| ../../../../../work-shared/llvm-project-source-18.1.0-r0/git/clang-tools-extra/clangd/index/dex/dexp/Dexp.cpp: In substitution of ‘template<class _Functor, class, class> std::function<std::unique_ptr<clang::clangd::{anonymous}::
Command>()>::function(_Functor) [with _Functor = <missing>; <template-parameter-1-2> = <missing>; <template-parameter-1-3> = <missing>]’:
| ../../../../../work-shared/llvm-project-source-18.1.0-r0/git/clang-tools-extra/clangd/index/dex/dexp/Dexp.cpp:372:1:   required from here
| ../../../../../work-shared/llvm-project-source-18.1.0-r0/git/clang-tools-extra/clangd/index/dex/dexp/Dexp.cpp:372:1: internal compiler error: Segmentation fault
|   372 | };
|       | ^
| Please submit a full bug report,
| with preprocessed source if appropriate.
| See <file:///usr/share/doc/gcc-10/README.Bugs> for instructions.
```
Nevertheless having the PACKAGECONFIG option available to simply turn compilation off in our builds is fine for me.

kraj (approved):
lgtm

lorenzph:
@kraj: Are the build logs of the CI run available somewhere (based on the log output it doesn't look like the CI failure is caused by this change but I'd like to make sure)?

kraj:
> @kraj: Are the build logs of the CI run available somewhere (based on the log output it doesn't look like the CI failure is caused by this change but I'd like to make sure)?

they are right in github actions link above. Click on 'details'

lorenzph:
> > @kraj: Are the build logs of the CI run available somewhere (based on the log output it doesn't look like the CI failure is caused by this change but I'd like to make sure)?
> 
> they are right in github actions link above. Click on 'details'

I meant the detailed logs of the bitbake run (or I am simply failing to find the right link :-)). The logs from https://github.com/kraj/meta-clang/actions/runs/7899722329/attempts/2?pr=912:
```
 Traceback (most recent call last):
  File "/home/khem/actions-runner-meta-clang/_work/meta-clang/meta-clang/yoe/sources/poky/meta/lib/oeqa/core/decorator/__init__.py", line 35, in wrapped_f
    return func(*args, **kwargs)
  File "/home/khem/actions-runner-meta-clang/_work/meta-clang/meta-clang/yoe/sources/poky/meta/lib/oeqa/runtime/cases/parselogs.py", line 185, in test_parselogs
    self.assertEqual(errcount, 0, msg=self.msg)
AssertionError: 1 != 0 : Log: /home/khem/actions-runner-meta-clang/_work/meta-clang/meta-clang/yoe/build/tmp/work/qemuarm64-yoe-linux/yoe-sdk-image/1.0/target_logs/postinstall.log
-----------------------
Central error: * opkg_cmd_exec: Command failed to capture privilege lock: Resource temporarily unavailable.
***********************
 * opkg_lock: Could not lock /run/opkg.lock: Resource temporarily unavailable.
 * opkg_cmd_exec: Command failed to capture privilege lock: Resource temporarily unavailable.
 ```
sounds like the failure is unrelated to change (but there is a chance that the issue leading up to this would be visible from the bitbake logs).

kraj:
> > > @kraj: Are the build logs of the CI run available somewhere (based on the log output it doesn't look like the CI failure is caused by this change but I'd like to make sure)?
> > 
> > 
> > they are right in github actions link above. Click on 'details'
> 
> I meant the detailed logs of the bitbake run (or I am simply failing to find the right link :-)). The logs from https://github.com/kraj/meta-clang/actions/runs/7899722329/attempts/2?pr=912:
> 
> ```
>  Traceback (most recent call last):
>   File "/home/khem/actions-runner-meta-clang/_work/meta-clang/meta-clang/yoe/sources/poky/meta/lib/oeqa/core/decorator/__init__.py", line 35, in wrapped_f
>     return func(*args, **kwargs)
>   File "/home/khem/actions-runner-meta-clang/_work/meta-clang/meta-clang/yoe/sources/poky/meta/lib/oeqa/runtime/cases/parselogs.py", line 185, in test_parselogs
>     self.assertEqual(errcount, 0, msg=self.msg)
> AssertionError: 1 != 0 : Log: /home/khem/actions-runner-meta-clang/_work/meta-clang/meta-clang/yoe/build/tmp/work/qemuarm64-yoe-linux/yoe-sdk-image/1.0/target_logs/postinstall.log
> -----------------------
> Central error: * opkg_cmd_exec: Command failed to capture privilege lock: Resource temporarily unavailable.
> ***********************
>  * opkg_lock: Could not lock /run/opkg.lock: Resource temporarily unavailable.
>  * opkg_cmd_exec: Command failed to capture privilege lock: Resource temporarily unavailable.
> ```
> 
> sounds like the failure is unrelated to change (but there is a chance that the issue leading up to this would be visible from the bitbake logs).

thats is bitbake run ( actually runtime logs  in qemuarm64 ), however, this failure is known, it happens when more than one qemu instance is running same test case, it happens rarely but happens. I have started another run and see if it goes through.

PR #924, opened 2024-03-19
layer.conf: Mark it scarthgap compatible

saininav:
Downstream projects need time to switch to opaque
pointers in llvm16 which requires in some cases
significant effort.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -7,7 +7,7 @@ BBFILES += "${LAYERDIR}/recipes*/*/*.bb ${LAYERDIR}/recipes*/*/*.bbappend"
 BBFILE_COLLECTIONS += "clang-layer"
 BBFILE_PATTERN_clang-layer := "^${LAYERDIR}/"
 BBFILE_PRIORITY_clang-layer = "7"
-LAYERSERIES_COMPAT_clang-layer = "mickledore nanbield"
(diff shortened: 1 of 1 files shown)


kraj (commented):
scarthgap is LTS and this would mean it will have to be supported for that long. is this something transitional ?

saininav:
Yes it is transitional. Once llvm16 support is available, we can remove it. 
https://github.com/intel/intel-graphics-compiler/issues/289

zboszor:
My alternative suggestion would be to introduce parallel versions of llvm in meta-clang to satisfy diverse software projects supporting specific llvm versions. E.g.:

- intel-graphics-compiler in meta-intel supports llvm 11-15
- python3-llvmlite in meta-python-ai supports llvm 14-15
- chipStar (soon to be introduced in meta-python-ai) is documented to support llvm 17, 15-16 mentioned as "maybe", and I tested that it at least builds with 18

FWIW, when using meta-clang/mickledore under scarthgap, I was able to build the Intel OpenCL runtime (a.k.a. intel-compute-runtime) but chromium-x11 (120.x) from meta-browser failed to build, even when pinned to the commit before the one that enabled Rust to build Chromium. Later versions need a newer clang version. This is a time when you can't satisfy everyone with a single LLVM version. YMMV.

PR #941, opened 2024-05-03
clang: use release tarball instead of git

Ecordonnier:
For clang 18.1.5, the llvm-project-source repository is about 3GB big, whereas the source tarball is 126MB big. Using the source tarball instead of the git repository speeds up the fetch step significantly.

Also, the git repository is getting so big that depending on the network setup, it is necessary to change git http.postBuffer options in order to avoid errors such as "fetch-pack: unexpected disconnect while reading sideband packet".

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):
seems ok. LGTM

PR #952, opened 2024-06-13
bpftrace: fix CVE-2024-2313

meenaligupta159:
If kernel headers need to be extracted, bpftrace will attempt to load them from a temporary directory. An unprivileged attacker could use this to force bcc to load compromised linux headers. Linux distributions which provide kernel headers by default are not affected by default.


kraj:
Thanks @meenaligupta159, i see the PR is against `master` branch is that correct ?
secondly, please squash both commits into one and for push to your branch

PR #954, opened 2024-06-16
bpftrace: fix CVE-2024-2313.patches via upload

meenaligupta159:
If kernel headers need to be extracted, bpftrace will attempt to load them from a temporary directory. An unprivileged attacker could use this to force bcc to load compromised linux headers. Linux distributions which provide kernel headers by default are not affected by default.

### DIFF
--- a/0001-bpftrace-fix-CVE-2024-2313.patch
+++ b/0001-bpftrace-fix-CVE-2024-2313.patch
@@ -1,0 +1,157 @@
+From f77ff583d0580ad57bb8742edad034b5877b1afe Mon Sep 17 00:00:00 2001
+From: Meenali Gupta <meenali.gupta@windriver.com>
+Date: Thu, 6 Jun 2024 04:37:47 +0000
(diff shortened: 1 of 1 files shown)


kraj (commented):
Please edit the commit message of the commit to reflect the changes correctly.

petermarko:
@kraj @meenaligupta159 this patch file is in root of the repository and obviously also not referenced in any recipe, so I'm not sure why it was merged

kraj:
> @kraj @meenaligupta159 this patch file is in root of the repository and obviously also not referenced in any recipe, so I'm not sure why it was merged

Yeah it is wrongly done I will have to revert it unless a PR to correct it is in place 

would you mind correcting it ?

petermarko:
I'll rather not handle patch which I don't know, so the PR author has to step in

kraj:
> I'll rather not handle patch which I don't know, so the PR author has to step in
fair enough. Thanks for reporting though, its huge help. I have reverted it via - https://github.com/kraj/meta-clang/pull/959

PR #972, opened 2024-07-08
Fix the build of bpftrace ptest failed

hongxu-jia:
…t failed

Due to commit [1][2], llvm-native was added to bpftrace to provide llvm-objcopy for the ptest.

But because of commit [3], llvm-native requires opengl in DISTRO_FEATURES to support libllvm.

In order to make the build of bpftrace ptest work, explicitly add libllvm to PACKAGECONFIG for llvm-native if ptest enabled.

According to [2][3], bpftrace used llvm-objcopy to generate BTF data for unit tests, regardless opengl in DISTRO_FEATURES, enable accelerated software renderer or not does not have side effect on bpftrace

[1] https://github.com/kraj/meta-clang/commit/bf6c02a47fb2ffec735df6c7dd642432f3681337
[2] https://github.com/bpftrace/bpftrace/commit/6cc531e35cd3bf77aa2a30cef3e3481c9c53042e
[3] https://git.openembedded.org/openembedded-core/commit/?id=ec22bfa67f6f1766102501d4593ce29aafe8c166

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/llvm/llvm_%.bbappend
+++ b/recipes-devtools/llvm/llvm_%.bbappend
@@ -1,0 +1,1 @@
+PACKAGECONFIG:append:class-native = " ${@bb.utils.contains('PTEST_ENABLED', '1', 'libllvm', '', d)}"


kraj (approved):
lgtm

kraj:
I had to revert this due to it causing check-layer to fail. meta-clang is yocto-compatible certified and we can not break that

PR #983, opened 2024-07-26
compiler-rt-sanitizers: Fix native build tooling paths

0xERIC:
Native builds of compiler-rt-sanitizers fail because llvm-ranlib, llvm-ar, and llvm-nm are not found where the recipe specifies that they'll be. Other recipes (in particular: compiler-rt) only specify the locations of those tools for target and nativesdk builds, and they build fine. This commit implements the same thing for compiler-rt-sanitizers, fixing native builds.

---
### Contributor checklist

### Reviewer Guidelines


0xERIC:
> is this needed on master as well ? if so please send a patch to master first and then backports to release branches can follow

It is. PR has been opened here: https://github.com/kraj/meta-clang/pull/984

PR #984, opened 2024-07-30
compiler-rt-sanitizers: Fix native build tooling paths

0xERIC:
Native builds of compiler-rt-sanitizers fail because llvm-ranlib, llvm-ar, and llvm-nm are not found where the recipe specifies that they'll be. Other recipes (in particular: compiler-rt) only specify the locations of those tools for target and nativesdk builds, and they build fine. This commit implements the same thing for compiler-rt-sanitizers, fixing native builds.

Note that there seems to be a separate issue with `compiler-rt-sanitizers` failing to build with `-DCOMPILER_RT_USE_BUILTINS_LIBRARY=ON` as is specified in the recipe. But with that set to `OFF`, the package fails to build as described above without this patch, and successfully builds with it applied.

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):
Lgtm

0xERIC:
Don't think I have write access, so you'll need to merge this (and the backport to kirkstone). I'll also do one for scarthgap and send that your way.

PR #987, opened 2024-07-30
clang: improve reproducibility

raburaibu:
Currently, class-target is reproducible, but class-nativesdk is not. What I did:

- noticed that BuildVariables.inc contains meaningless path fragments after running sed, so I adjusted sed commands to remove the rest of paths as well;
- moved common code into a function, which is then called twice;
- changed do_compile:prepend into do_configure:append, this is more conventional and intuitive;
- verified that the resulting code works for both target and nativesdk after these changes.

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):
thanks for doing this LGTM

PR #992, opened 2024-08-06
libclc: use cmake-qemu to clean up recipe

rossburton:
The cmake-qemu class handles using qemu-user as a CMake emulator, so use that and clean up the recipe.


kraj (approved):
lgtm

Issue #998, opened 2024-08-13
Consider move `TUNE_CCARGS:remove = "-mbranch-protection=standard"` to conf file


Oiouuu:
Hi, why not place the changes to TUNE_CCARGS in the conf file? 
This change should affect the global CFLAGS. If it's only set in `libcxx_git.bb`, other packages won’t be affected. 
I found that when building other packages, the cxx exception feature doesn’t work properly on aarch64. 
It only works if 
TUNE_CCARGS:remove:toolchain-clang = "${@bb.utils.contains('TUNE_FEATURES', 'aarch64', '-mbranch-protection=standard', '', d)}" 
is added globally.

kraj:
https://github.com/kraj/meta-clang/issues/963 reported this issue and fixing C++ runtime was enough to solve that case. However if you are seeing more issues with it then we might want to use it globally infact for all the distros using clang + gcc to build packages will show similar issue on arm64.  Can you add a package where this happens and removing it helps ?

Oiouuu:
Here is an example [vsomip](https://github.com/COVESA/vsomeip/blob/master/implementation/configuration/src/configuration_impl.cpp). You can see that Boost is used to load the configuration file with 'try-catch'. Sure I can remove `TUNE_FEATURES` in [vsomeip.bb](https://layers.openembedded.org/layerindex/recipe/401163/), but it's a runtime issue so I can't test all packages.

quaresmajose:
I am seeing the same in my distro.
I thought https://github.com/kraj/meta-clang/pull/968 would solve the problem but it didn't and we continue to have problems with the exceptions in runtime.
I will also test https://github.com/kraj/meta-clang/pull/999 to see the result.

We had to revert again https://github.com/foundriesio/meta-lmp/pull/1500 and apply our fix globally 
```
TUNE_CCARGS:remove:toolchain-clang = "${@bb.utils.contains('TUNE_FEATURES', 'aarch64', '-mbranch-protection=standard', '', d)}"
```

Oiouuu:
> I am seeing the same in my distro. I thought #968 would solve the problem but it didn't and we continue to have problems with the exceptions in runtime. I will also test #999 to see the result.
> 
> We had to revert again [foundriesio/meta-lmp#1500](https://github.com/foundriesio/meta-lmp/pull/1500) and apply our fix globally
> 
> ```
[...]

PR #999, opened 2024-08-15
Update libcxx_git.bb

Oiouuu:
Fix syntax error, just a little issue '-DD' -> '-D'

---
### Contributor checklist
- [x] Changes have been tested
- [x] `Signed-off-by` is present
- [x] The PR complies with the [Open Embedded Commit Patch Message Guidelines](http://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines)

PR #1000, opened 2024-08-20
layer.conf: remove '-mbranch-protection=standard' globally on aarch64

quaresmajose:
On aarch64 the handling of runtime exception is broken when build with clang.
There is currently a fix for libcxx on the layer but it is not enough and this same fix needs to be applied globally.

First fix attempt just include libcxx:
https://github.com/kraj/meta-clang/issues/963
https://github.com/kraj/meta-clang/pull/968

Remaining reported issue:
https://github.com/kraj/meta-clang/issues/998

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/conf/nonclangable.conf
+++ b/conf/nonclangable.conf
@@ -1,3 +1,6 @@
+# introduced in oe-core 8905639d breaks the handling of runtime exception when build with clang
+TUNE_CCARGS:remove:toolchain-clang = "${@bb.utils.contains('TUNE_FEATURES', 'aarch64', '-mbranch-protection=standard', '', d)}"
+
 # https://github.com/llvm/llvm-project/issues/78337
 TOOLCHAIN:pn-kernel-selftest = "gcc"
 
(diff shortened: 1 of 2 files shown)


conf/layer.conf
```diff
@@ -35,5 +35,8 @@ CLANGSDK ??= "0"
 
 LLVMVERSION = "18.1.8"
 
+# introduced in oe-core 8905639d breaks the handling of runtime exception when build with clang
+TUNE_CCARGS:remove:toolchain-clang = "${@bb.utils.contains('TUNE_FEATURES', 'aarch64', '-mbranch-protection=standard', '', d)}"
+
```
kraj, line 40: lets move this to `nonclangable.conf`

kraj (changes requested):

conf/layer.conf
kraj: and check if https://github.com/kraj/meta-clang/pull/999 made any difference

quaresmajose: I have tested it and unfortunately don't solve the issue, also reported that on https://github.com/kraj/meta-clang/issues/998

kraj (approved):
LGTM

PR #1012, opened 2024-10-01
u-boot: u-boot is clangable

jonmason:
Since clang is at version 19 now, remove the blocking of building u-boot.  I'm not seeing any issues with building u-boot and running testimage for qemuarm, qemuarm64, and qemumips.
Based on the original comment, this limitation probably should've been only applied to mips.

Leaving the BSP versions of u-boot in nonclangable, as I don't know if those have issues or not, and thought it better to err on the side of caution.


kraj (approved):
Lgtm we will add exceptions as they become evident

PR #1020, opened 2024-10-16
ostree: Use libcc for compiler runtime when using glibc for system

kraj:
glibc static library expects builtins from libgcc which may not exist in compiler-rt, therefore use libgcc explicitly.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/conf/nonclangable.conf
+++ b/conf/nonclangable.conf
@@ -370,7 +370,6 @@ COMPILER_RT:remove:pn-webkitgtk:powerpc = "-rtlib=compiler-rt"
 COMPILER_RT:remove:pn-m4:armeb = "-rtlib=compiler-rt"
 COMPILER_RT:remove:pn-ruby:armeb = "-rtlib=compiler-rt"
 COMPILER_RT:remove:pn-webkitgtk:armeb = "-rtlib=compiler-rt"
-
 # build/lib/libQt5Widgets.so: undefined reference to `__lshrti3'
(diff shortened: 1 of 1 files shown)


conf/nonclangable.conf
```diff
@@ -398,6 +397,7 @@ COMPILER_RT:pn-mpich:toolchain-clang:x86-64 = "-rtlib=libgcc --unwindlib=libgcc"
 COMPILER_RT:pn-aufs-util:libc-glibc:toolchain-clang:x86-64 = "-rtlib=libgcc --unwindlib=libgcc"
 COMPILER_RT:pn-libhugetlbfs:libc-glibc:toolchain-clang:x86-64 = "-rtlib=libgcc --unwindlib=libgcc"
 COMPILER_RT:pn-tsocks:libc-glibc:toolchain-clang:x86-64 = "-rtlib=libgcc --unwindlib=libgcc"
+COMPILER_RT:pn-ostree:toolchain-clang:libc-glibc = "-rtlib=libgcc --unwindlib=libgcc"
```
quaresmajose, line 400: I have used and we only need it for x86-64, for reference I https://github.com/foundriesio/meta-lmp/commit/0abb08c83aee7027711f538e29a5ff16dc334304

kraj: nice !. Since you have done this and perhaps tested it a bit longer, would you mind proposing it here ? I would like to include those changes instead

PR #1037, opened 2024-12-02
clang: remove PROVIDES:append settings

ChenQi1989:
With these settings, we'll have multiple providers of llvm. This means that with pristine poky master + meta-clang master, we'll have some output like below:

  NOTE: Multiple providers are available for llvm-native (llvm-native, clang-native)
  Consider defining a PREFERRED_PROVIDER entry to match llvm-native

And when running 'bitbake world', we'll get error messages like below:

  ERROR: Multiple .bb files are due to be built which each provide llvm-native:
    virtual:native:/PATH/TO/poky/meta/recipes-devtools/llvm/llvm_19.1.4.bb
    virtual:native:/PATH/TO/Yocto/poky/meta-clang/recipes-devtools/clang/clang_git.bb

As the README.md has already specified how to use this layer, let's remove such settings before sorting things out.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -303,9 +303,6 @@ do_install:append:class-nativesdk () {
     sed -i -e 's,${B},,g' ${D}${libdir}/cmake/llvm/LLVMConfig.cmake
 }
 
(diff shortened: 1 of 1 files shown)


kraj (approved):
lgtm

Issue #1061, opened 2025-03-14
Commit f002eb5a breaks our build


aprospero:
### Prerequisites

Yocto Build on Scarthgap, sdk build for x86_64 - crossplatform ARM cortex a9.

### Symptoms

The task do_install for recipe nativesdk-clang fails with an error:

```
| sed: can't read /yoctolinux/build/tmp/work/x86_64-nativesdk-sdk-linux/nativesdk-clang/18.1.6/image/opt/xxx/cortexa9t2hf-neon-scarthgap/sysroots/x86_64-sdk-linux/usr/lib/cmake/llvm/LLVMConfig.cmake: No such file or directory

| WARNING: exit code 2 from a shell command.

ERROR: Task (virtual:nativesdk:yoctolinux/artifacts/github.com/kraj/meta-clang/recipes-devtools/clang/clang_git.bb:do_install) failed with exit code '1'
```
### Reproduction

vanilla build on current scarthgap (commit 8c77b427) 

with 

EXTRA_OECMAKE:append:pn-nativesdk-clang = " -DLLVM_INSTALL_TOOLCHAIN_ONLY=ON"

as suggested here: 

https://llvm.org/docs/BuildingADistribution.html

### Expected behavior
No error. ;)

### Findings

Commit f002eb5a doesn't seem to take into account that LLVMConfig.cmake doesn't exist - or the creation of LLVMConfig.cmake broke silently.

kraj:
Thanks for report, it can be fixed by checking the .cmake file for existing before operating on it.

```
if [ -e ${D}${libdir}/cmake/llvm/LLVMConfig.cmake ]
    sed -i -e 's,${B},,g' ${D}${libdir}/cmake/llvm/LLVMConfig.cmake
fi
```

would be ok. Can you test this fix and if it works for you then send a PR against `master` ( if master needs it ) and `scarthgap`

PR #1063, opened 2025-03-17
fix: [Issue #1061] - check for existence before trying to process a file

aprospero:
### Abstract
In certain configurations LLVMConfig.cmake does not exist in the build output. If that's the case do_install breaks for the nativesdk-clang recipe.

### Reproduce
To reproduce set the following on a vanilla build:

EXTRA_OECMAKE:append:pn-nativesdk-clang = " -DLLVM_INSTALL_TOOLCHAIN_ONLY=ON"

as suggested here:

https://llvm.org/docs/BuildingADistribution.html

### Remarks
The corresponding issue #1061 blames commit f002eb5a but 716d7efb is also to blame and it is not ruled out that the problem could affect other configurations and other tasks.

**Question**: 716d7efb is from Oct. 2023 - so for a while out in the wild. But nobody seems to have similar problems like we have. If we remove the EXTRA_OECMAKE:append the build works fine. Does noone actually builds clang with the suggestd EXTRA_OECMAKE:append, or why is it that we seem to be the only ones to be affected?

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -284,7 +284,9 @@ do_install:append:class-native () {
 }
 
 do_install:append:class-nativesdk () {
-    sed -i -e "s|${B}/./bin/||g" ${D}${libdir}/cmake/llvm/LLVMConfig.cmake
+    if [ -e ${D}${libdir}/cmake/llvm/LLVMConfig.cmake ] ; then
+        sed -i -e "s|${B}/./bin/||g" ${D}${libdir}/cmake/llvm/LLVMConfig.cmake
+    fi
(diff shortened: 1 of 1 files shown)


kraj (commented):
LGTM

kraj:
@shr-project since these changes being modified here came from you, can you also take a look and let me know if its ok.

shr-project:
I've only cherry-picked that reproduciblity fix from master branch to scarthgap and indeed I don't use `EXTRA_OECMAKE:append:pn-nativesdk-clang = " -DLLVM_INSTALL_TOOLCHAIN_ONLY=ON"` in my builds.

LGTM

PR #1071, opened 2025-03-31
compiler-rt: fix do_install error

kkang-wr:
After commit

* 6da0aba clang.bbclass,compiler-rt: Do not set -target for native clang

it installs files to ${nonarch_libdir}/linux rather than ${libdir}/linux when multilib enabled. Update to fix following do_install error:

mv: cannot stat '/path_to/tmp/work/core2-64-poky-linux/compiler-rt/20.1.1/image/usr/lib64/linux': No such file or directory

---
### Contributor checklist

### Reviewer Guidelines


kraj (approved):

shr-project:
What about removing the whole conditional? That's what I was using with https://github.com/kraj/meta-clang/pull/1072

kraj:
> What about removing the whole conditional? That's what I was using with #1072

yeah thats perhaps better to do.

kraj:
fixed with  #1072

PR #1075, opened 2025-04-03
More fixes for meta-clang vs OE-Core

lumag:
Several more patches for meta-clang, resolving sysroot-native conflict when building Mesa with libclc

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang/0033-llvm-Add-libunwind.pc.in-and-llvm-config-scripts.patch
+++ b/recipes-devtools/clang/clang/0033-llvm-Add-libunwind.pc.in-and-llvm-config-scripts.patch
@@ -35,7 +35,7 @@ new file mode 100644
 index 000000000000..6a0dd54b8eab
 --- /dev/null
 +++ b/llvm/tools/llvm-config/llvm-config
-@@ -0,0 +1,52 @@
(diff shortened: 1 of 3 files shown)


recipes-devtools/llvm/llvm_%.bbappend
```diff
@@ -0,0 +1,2 @@
+# Temporarily until a corresponding change is merged to OE-Core
```
quaresmajose, line 1: Since this is temporary, it would be preferable to use the specific version `llvm_20.1.0.bbappend` so that this change is not forgotten here.

lumag: I tried doing this. Unfortunately, it makes handling of this PR much more complicated as the Yoe is not in full sync with OE-Core. I tried pushing a correct versioned bbappend and the build failed. Anyway, I've opened #1076 as a draft, so that we don't forget to drop it.

quaresmajose: yeah the draft is a good reminder.

lumag:
Mesa pull request has been merged to OE-Core, so there is no more need for llvm bbappend. Dropped it from this pull request.

PR #1085, opened 2025-04-16
Fixes for OE-Core compatibility

lumag:
Note: I wasn't able to build nativesdk-clang, it can not find gcc installation correctly and thus it can't provide paths to crtbeginS.o / crtendS.o /-lgcc to linker.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang_git.bb
+++ b/recipes-devtools/clang/clang_git.bb
@@ -214,8 +214,9 @@ EXTRA_OECMAKE:append:class-target = "\
 "
 
 DEPENDS = "binutils zlib zstd libffi libxml2 libxml2-native ninja-native swig-native"
-DEPENDS:append:class-nativesdk = " clang-crosssdk-${SDK_SYS} virtual/nativesdk-cross-binutils nativesdk-python3"
-DEPENDS:append:class-target = " clang-cross-${TARGET_ARCH} python3 compiler-rt libcxx"
+DEPENDS:append:class-native = "${@['llvm-native', '']['llvm' in (d.getVar('PROVIDES') or '')]}"
+DEPENDS:append:class-nativesdk = " clang-crosssdk-${SDK_SYS} virtual/nativesdk-cross-binutils nativesdk-python3 ${@['nativesdk-llvm', '']['llvm' in (d.getVar('PROVIDES') or '')]}"
+DEPENDS:append:class-target = " clang-cross-${TARGET_ARCH} python3 compiler-rt libcxx ${@['llvm', '']['llvm' in (d.getVar('PROVIDES') or '')]}"
 
 RRECOMMENDS:${PN} = "binutils"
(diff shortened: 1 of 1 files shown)


recipes-devtools/clang/clang_git.bb
```diff
@@ -214,8 +214,9 @@ EXTRA_OECMAKE:append:class-target = "\
 "
 
 DEPENDS = "binutils zlib zstd libffi libxml2 libxml2-native ninja-native swig-native"
-DEPENDS:append:class-nativesdk = " clang-crosssdk-${SDK_SYS} virtual/nativesdk-cross-binutils nativesdk-python3"
-DEPENDS:append:class-target = " clang-cross-${TARGET_ARCH} python3 compiler-rt libcxx"
+DEPENDS:append:class-native = "llvm-native"
```
kraj, line 217: this is not going to work in all case, because clang-native can also provide llvm-native, it will end up with dep loops then

recipes-devtools/clang/clang_git.bb
```diff
@@ -484,9 +484,12 @@ SYSROOT_DIRS:append:class-target = " ${nonarch_libdir}"
 SYSROOT_PREPROCESS_FUNCS:append:class-target = " clang_sysroot_preprocess"
 
 clang_sysroot_preprocess() {
-	install -d ${SYSROOT_DESTDIR}${bindir_crossscripts}/
-	install -m 0755 ${S}/llvm/tools/llvm-config/llvm-config ${SYSROOT_DESTDIR}${bindir_crossscripts}/
-	ln -sf llvm-config ${SYSROOT_DESTDIR}${bindir_crossscripts}/llvm-config${PV}
+	if ${@ 'true' if 'llvm' in d.getVar('PROVIDES') or '' else 'false' } ; then
+		install -d ${SYSROOT_DESTDIR}${bindir_crossscripts}/
+		install -m 0755 ${S}/llvm/tools/llvm-config/llvm-config ${SYSROOT_DESTDIR}${bindir_crossscripts}/
+		ln -sf llvm-config ${SYSROOT_DESTDIR}${bindir_crossscripts}/llvm-config${PV}
```
kraj, line 490: this is also going to be problematic when llvm bits are provided by clang-native

kraj (commented):
thanks for trying to fix this, while this will build default case it will break a case where clang is providing LLVM bits, we can switch to using the llvm provided by core but it does not have all the patches that meta-clang applies

recipes-devtools/clang/clang_git.bb
lumag: fixed in rhe new version

recipes-devtools/clang/clang_git.bb
lumag: But there is a PROVIDES check. Isn't it enough?

PR #1087, opened 2025-04-22
clang: Update dependencies for cross-compilation support

psaavedra:
Modify `clang_git.bb`, `clang-cross-canadian_git.bb` and `clang-crosssdk_git.bb` to conditionally depend on `virtual/${HOST_PREFIX}binutils` or `virtual/${TARGET_PREFIX}binutils` for `scarthgap` and `styhead` layers.

This prevents `Nothing PROVIDES 'virtual/nativesdk-cross-binutils'` in old but still compatible layers as `scarthgap` and `styhead`.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/recipes-devtools/clang/clang-cross-canadian_git.bb
+++ b/recipes-devtools/clang/clang-cross-canadian_git.bb
@@ -12,7 +12,7 @@ require clang.inc
 require common-source.inc
 inherit cross-canadian
 
-DEPENDS += "nativesdk-clang binutils-cross-canadian-${TRANSLATED_TARGET_ARCH} virtual/nativesdk-cross-binutils virtual/nativesdk-libc"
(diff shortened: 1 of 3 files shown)


kraj:
@psaavedra we have specific branches for supporting styhead and scarthgap why do we need this change in master ?

psaavedra:
> @psaavedra we have specific branches for supporting styhead and scarthgap why do we need this change in master ?

I am using  `master` branch because it claims compatibility with the `scarthgap` and `styhead` Yocto releases (ref: https://github.com/kraj/meta-clang/blob/master/conf/layer.conf#L10).

If that is not the expected behavior, then it would be better to mark the branch as compatible only with `walnascar`. In my particular case, using the `master` branch facilitates the compilation of a backport of the mesa 25.0.2 recipe in a `scarthgap` Yocto environment.

psaavedra:
This PR is not complete. I have recently discovered some issues compiling other
components. I don't think these changes are worth adding. Let's close this PR.

PR #1116, opened 2025-07-18
Fix YP compatibility check

kraj:
---
### Contributor checklist

### Reviewer Guidelines


shr-project (approved):
These were also causing failures for builds with target toolchain gcc as reported on #yocto 2025-07-14

23:17 < JaMa> khem: there is small issue with rust (and rust-llvm) when building with meta-clang, the bbappend there sets TOOLCHAIN_NATIVE to clang, so clang-native is inheritted (even for target builds) and target builds with gcc fail, because they don't recognize --rtlib=libgcc --unwindlib=libgcc when target CC is gcc-cross, I guess you're not seeing this because most builds with meta-clang will use clang also for target
23:17 < JaMa> builds
23:17 < JaMa> not sure what's the correct fix, should clang-native append LDFLAGS only for class-native?

So good I won't need to BBMASK them anymore, thanks :).

PR #1127, opened 2025-08-29
lxc: Use gcc instead of clang compiler

Deepak-Rathore580:
While compiling lxc with clang, it fails with compilation error and not compatible with clang compiler. Hence, updated TOOLCHAIN of lxc to use gcc.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/conf/nonclangable.conf
+++ b/conf/nonclangable.conf
@@ -202,6 +202,7 @@ TOOLCHAIN:pn-gcc-for-nvcc-runtime:tegra = "gcc"
 TOOLCHAIN:pn-optee-os:imx-nxp-bsp = "gcc"
 
 # Does not compile with clang 18
+TOOLCHAIN:pn-bpftool = "gcc"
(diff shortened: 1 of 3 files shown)


conf/nonclangable.conf
```diff
@@ -201,6 +201,10 @@ TOOLCHAIN:pn-gcc-for-nvcc-runtime:tegra = "gcc"
 #|       | ^
 TOOLCHAIN:pn-optee-os:imx-nxp-bsp = "gcc"
 
+# Does not compile with clang 18
+# ERROR: LLVM's thinLTO only works with gnu gold, lld, lld-link, and ld64, not ld.bfd
+TOOLCHAIN:pn-lxc = "gcc"
```
kraj, line 206: we do not enable thin-lto by default, but its a possible DISTRO_FEATURE so it might be better to check for this in DISTRO_FEATURES and then resort to always using gcc toolchain

kraj (changes requested):

kraj (approved):
I will apply it for now.

PR #1128, opened 2025-08-29
bpftool: Use gcc instead of clang compiler

Deepak-Rathore580:
While compiling bpftool with clang, it fails with below error: | error: no member named 'id' in 'struct bpf_link'
|           return BPF_CORE_READ((struct bpf_link *)ent, id);

bpftool is kernel tool to inspect and manipulate eBPF programs and maps. And kernel is built using GCC by default. Maintaining compatibility with Clang is an ongoing challenge, and using Clang solely for building bpftool is ineffective unless the entire kernel is compiled with Clang.

Hence, updated TOOLCHAIN of bpftool to gcc.

---
### Contributor checklist

### Reviewer Guidelines

### DIFF
--- a/conf/nonclangable.conf
+++ b/conf/nonclangable.conf
@@ -201,6 +201,9 @@ TOOLCHAIN:pn-gcc-for-nvcc-runtime:tegra = "gcc"
 #|       | ^
 TOOLCHAIN:pn-optee-os:imx-nxp-bsp = "gcc"
 
+# Does not compile with clang 18
+TOOLCHAIN:pn-bpftool = "gcc"
+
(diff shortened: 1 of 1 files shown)


conf/nonclangable.conf
```diff
@@ -201,6 +201,9 @@ TOOLCHAIN:pn-gcc-for-nvcc-runtime:tegra = "gcc"
 #|       | ^
 TOOLCHAIN:pn-optee-os:imx-nxp-bsp = "gcc"
 
+# Does not compile with clang 18
+TOOLCHAIN:pn-bpftool = "gcc"
```
kraj, line 205: Its an independent tool and infact should work well with clang, please fix the code.

kraj (changes requested):

PR #1163, opened 2026-06-09
clang: Add status of CVE-2024-7883

divyanshur14:
The patch (CVE-2024-7883.patch) is correctly applied via llvm-project-source's do_patch task. However, because common-source.inc sets SRC_URI = "" in the clang recipe (shared-source architecture), cve_check finds no patches in the clang recipe and reports the CVE as Unpatched.

Add an explicit CVE_STATUS declaration to mark the CVE as fixed and document that the patch is applied via llvm-project-source, without altering the shared-source build flow.

CVE: CVE-2024-7883

Signed-off-by: Divyanshu Rathore <Divyanshu.DR.Rathore@bti.bmwgroup.com>


kraj (approved):
lgtm

PR #1165, opened 2026-06-25
conf/layer.conf: remove unused BBFILES_DYNAMIC entries

DeepeshWR:
Remove stale dynamic layer references (browser-layer, networking-layer, meta-python, qt5-layer) from BBFILES_DYNAMIC in layer.conf.

These directories no longer exist under dynamic-layers/, so the patterns are dead code and have no effect.

(cherry picked from commit 0baedd06e78d26d3b927471cfd91960b7229a178)


kraj (approved):
LGTM

